PR #27, opened 2025-03-27
Improve Feedback Handling for Synchronous Replication and WAL message 'w' flood control.

ibrarahmad:
This PR introduces two key enhancements to improve synchronous replication behavior and WAL sender efficiency:

1 - Enhanced Synchronous Replication Feedback
Previously, when a synchronous replica was attached, the spock set synchronous_commit to off, potentially sending feedback prematurely before receiving acknowledgment from the remote synchronous standby. Introduced tracking of locally committed LSNs. Feedback is now held back until an acknowledgment from the synchronous replica is received. This change improves data consistency by ensuring durability before transaction acknowledgment.

2 - Flood Control for 'w' Messages
We are now sending the feedback if 1/2 the wal_sender_timeout occurs or in case of 10 'w' messages to avoid the wal_sender_timeout on the other side in case of too many 'w' messages.

### DIFF
--- a/spock_apply.c
+++ b/spock_apply.c
@@ -1870,7 +1870,6 @@ handle_startup(StringInfo s)
 static void
 spock_apply_worker_on_exit(int code, Datum arg)
 {
-	/* FIXME: Do we really need to flush the sync_replica_lsn*/
 	spock_apply_worker_detach();
 }
 
--- a/spock_apply.c
+++ b/spock_apply.c
@@ -2855,8 +2855,9 @@ apply_work(PGconn *streamConn)
 
 				if (c == 'w')
 				{
-					XLogRecPtr	start_lsn;
-					XLogRecPtr	end_lsn;
+					XLogRecPtr start_lsn;
+					XLogRecPtr end_lsn;
+					static int w_message_count = 0;
 
 					start_lsn = pq_getmsgint64(&s);
 					end_lsn = pq_getmsgint64(&s);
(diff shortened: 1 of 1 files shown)


spock_apply.c
```diff
@@ -2752,6 +2869,24 @@ apply_work(PGconn *streamConn)
 					if (last_received < end_lsn)
 						last_received = end_lsn;
 
+					w_message_count++;
+
+					/*
+					 * Send feedback if wal_sender_timeout/2 has passed or after 10 'w' messages.
+					 */
+					if (TimestampDifferenceExceeds(last_receive_timestamp, GetCurrentTimestamp(), wal_sender_timeout / 2) ||
```
hqakhtar, line 2882: GetCurrentTimestamp() is being called repeatedly in this code block. We should only call it once and save it in a local variable.

spock_apply.c
```diff
@@ -2504,23 +2532,110 @@ get_flush_position(XLogRecPtr *write, XLogRecPtr *flush)
...
+{
+	dlist_mutable_iter iter1;
+	RemoteSyncPosition *syncpos;
+
+	Assert(WalSndCtl->sync_standbys_defined);
+	if (dlist_is_empty(&sync_replica_lsn))
+		return;
+
+	/* Acquire lock to update the sync position */
+	LWLockAcquire(SyncRepLock, LW_EXCLUSIVE);
```
hqakhtar, line 2595: Do we actually need to acquire this lock?

spock_apply.c
```diff
@@ -2504,23 +2532,110 @@ get_flush_position(XLogRecPtr *write, XLogRecPtr *flush)
...
+ * This ensures that we only send feedback that is committed and acknowledged
+ * by the synchronous replica.
+ */
+static void
+get_feedback_position(XLogRecPtr *recvpos, XLogRecPtr *writepos, XLogRecPtr *flushpos, XLogRecPtr *max_recvpos)
+{
+	dlist_mutable_iter iter1;
+	RemoteSyncPosition *syncpos;
+
+	Assert(WalSndCtl->sync_standbys_defined);
```
hqakhtar, line 2590: WalSndCtl should be assigned to a local volatile function variable.

spock_apply.c
```diff
@@ -1843,6 +1870,7 @@ handle_startup(StringInfo s)
 static void
 spock_apply_worker_on_exit(int code, Datum arg)
 {
+	/* FIXME: Do we really need to flush the sync_replica_lsn*/
```
hqakhtar: We should perform the cleanup on shared memory exit to ensure that:
(1) we are sending back the feedback confirmation to the publisher,
(2) cleaning up the list.

spock_apply.c
ibrarahmad: Yes, we need that lock; we copied that code from PostgreSQL core. Later, we will see that we can have a better approach.

PR #32, opened 2025-04-08
Improve handling of JSONB 'info' field extraction in `node_fromtuple`

ibrarahmad:
This patch refactors the extraction of the "tiebreaker" value from the `info` JSONB field in the `node_fromtuple` function. Instead of using `DirectFunctionCall2`, it now sets up and invokes the `jsonb_object_field_text` function call explicitly using `FunctionCallInfoData`. This approach provides better control over function invocation and error handling.

### DIFF
--- a/spock_node.c
+++ b/spock_node.c
@@ -36,6 +36,9 @@
 #include "utils/lsyscache.h"
 #include "utils/pg_lsn.h"
 #include "utils/rel.h"
+#include "fmgr.h"
+#include "funcapi.h"
+
 
 #include "spock_node.h"
 #include "spock_repset.h"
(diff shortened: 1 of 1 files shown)


rasifr (changes requested):
It looks like we're retrieving the "tiebreaker" field from a jsonb column, which can potentially return NULL. The current code uses DirectFunctionCall2(), which does not support functions that can return NULL and will raise an error if that happens. 

To fix this, we should replace the DirectFunctionCall2() call with OidFunctionCall2(F_JSONB_OBJECT_FIELD_TEXT, ...), which is NULL-safe.

PR #36, opened 2025-04-15
The local tuple for the exception was being created in a no-longer-available memory context

hqakhtar:
Local tuple was being saved in a memory context that was getting cleaned up as the transaction was getting rolled back or released.
    
So, perserving local tuple in the MessageContext ensures that it is available beyond the scope of the subtranction.


rasifr (approved):
Yeah this issue was previously identified and resolved in a different branch, but it should have gone to main first. Correcting the memory context addresses the root cause, and the changes in this patch are consistent with the fix applied earlier. This should fully resolve the issue.

PR #38, opened 2025-04-25
Zero downtime add node

wieck:
These commits provide the ability to create a subscription in disabled state and later forward the corresponding logical replication slot to a given commit timestamp. This is the Spock side functionality needed to build tooling that can expand a Spock cluster beyond two nodes without any application downtime.


ibrarahmad (approved):
LGTM

PR #43, opened 2025-05-13
UC Insert Exists

rasifr:
Introduce a new GUC `spock.check_all_uc_indexes` (default: false) that, when enabled, allows INSERT conflict detection to fall back on all usable unique indexes (excluding PK and replica identity indexes) if primary and replica identity indexes fail to find a match.

This feature enhances conflict resolution reliability in cases where unique constraints exist but are not part of PK/RI.

- Added new GUC spock.check_all_uc_indexes
- Implemented FindReplTupleByUCIndex() to scan all valid unique indexes
- Refactored index conflict checking logic into reusable helpers:
  - IsIndexUsableForInsertConflict()
  - SpockRelationFindReplTupleByIndex()
- Changed conflict type from CONFLICT_INSERT_INSERT to CONFLICT_INSERT_EXISTS for better semantics

### DIFF
--- a/spock.c
+++ b/spock.c
@@ -130,6 +130,7 @@ bool	spock_include_ddl_repset = false;
 bool	allow_ddl_from_functions = false;
 int		restart_delay_default;
 int		restart_delay_on_exception;
+bool	check_all_uc_indexes = false;
 
 
 void _PG_init(void);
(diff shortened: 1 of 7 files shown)


wieck:
Passes functional check on all supported PG versions. Code looks great.

PR #44, opened 2025-05-13
Change current devel version from 4.1.0 to 5.0

wieck:
The current development branch main is not network protocol compatible with the production branch 4.0.x. This will require a cluster wide (big bang) upgrade of all nodes at once. No way of upgrading Spock in a rolling fashion. Because of that we decided to bump the version number up to 5.0.

At the same time we change the numbering scheme to the pgEdge version policy.


rasifr (approved):
The changes looks good to me, the wait sync functionality works. so does extension upgrade.

PR #45, opened 2025-05-20
Exception sub_disable behaviour

rasifr:


spock_apply.c
```diff
@@ -2899,6 +2874,12 @@ apply_work(PGconn *streamConn)
 					send_feedback(applyconn, last_received, GetCurrentTimestamp(),
 								  false);
 				}
+
+				/* reset replication session to avoid reuse of it after error. */
+				if (MyApplyWorker->use_try_block &&
+					(exception_behaviour == TRANSDISCARD ||
+					 exception_behaviour == SUB_DISABLE))
+					replorigin_session_reset();
```
rasifr: On second thought, this isn't necessary. If control has reached this block, it means the transaction was rolled back without an error being thrown. Since no error occurred, there's no need to reset the replication state. If there is no other issue, I will update this change.

hqakhtar (approved):
LGTM

PR #49, opened 2025-05-28
Enhance synchronous replication feedback handling to avoid premature acknowledgment.

ibrarahmad:
When a synchronous replica is attached, the system had previously set synchronous_commit to 'off', which could lead to premature feedback being sent before receiving an acknowledgment from the remote synchronous replica. This behavior posed a risk to data consistency.

To address this, a list of locally committed LSNs is maintained. Feedback is now delayed until acknowledgment is received from the remote synchronous replica, ensuring transaction durability while preventing unnecessary blocking.

Now the feedback will be sent if 1/2 the wal_sender_timeout occurs or in case of 10 'w' messages.

### DIFF
--- a/spock_apply.c
+++ b/spock_apply.c
@@ -2529,7 +2529,7 @@ append_feedback_position(XLogRecPtr recvpos)
 	RemoteSyncPosition *syncpos;
 	MemoryContext oldctx;
 
-	Assert(WalSndCtl->sync_standbys_defined);
+	Assert(WalSndCtl->sync_standbys_status & SYNC_STANDBY_DEFINED);
 
 	if (get_flush_position(&writepos, &flushpos))
 	{
@@ -2570,7 +2570,7 @@ get_feedback_position(XLogRecPtr *recvpos, XLogRecPtr *writepos, XLogRecPtr *flu
 	dlist_mutable_iter iter1;
 	RemoteSyncPosition *syncpos;
 
-	Assert(WalSndCtl->sync_standbys_defined);
+	Assert(WalSndCtl->sync_standbys_status & SYNC_STANDBY_DEFINED);
 	if (dlist_is_empty(&sync_replica_lsn))
 		return;
 
--- a/spock_apply.c
+++ b/spock_apply.c
@@ -71,6 +71,8 @@
 #include "utils/memutils.h"
 #include "utils/pg_lsn.h"
 #include "utils/snapmgr.h"
+#include "replication/syncrep.h"
+#include "replication/walsender_private.h"
 
 #include "spock_common.h"
 #include "spock_conflict.h"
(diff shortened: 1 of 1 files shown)


spock_apply.c
```diff
@@ -2487,23 +2514,116 @@ get_flush_position(XLogRecPtr *write, XLogRecPtr *flush)
...
+ * This ensures that we only send feedback that is committed and acknowledged
+ * by the synchronous replica.
+ */
+static void
+get_feedback_position(XLogRecPtr *recvpos, XLogRecPtr *writepos, XLogRecPtr *flushpos, XLogRecPtr *max_recvpos)
+{
+	dlist_mutable_iter iter1;
+	RemoteSyncPosition *syncpos;
+
+	Assert(WalSndCtl->sync_standbys_defined);
```
rasifr: This needs to be updated

spock_apply.c
```diff
@@ -994,6 +1018,9 @@ handle_commit(StringInfo s)
 
 		CommitTransactionCommand();
 
+		if (WalSndCtl->sync_standbys_status & SYNC_STANDBY_DEFINED)
```
rasifr, line 1021: Perhaps we should consider relying on synchronous_standby_names by using GetConfigOptionByName("synchronous_standby_names", ...) instead of directly accessing WalSndCtl->sync_standbys_status. This way, we wouldn't have to restrict Spock to work only with newer PostgreSQL versions.

bonesmoses: Wouldn't it have to be both? The variable could be misconfigured or the other nodes unreachable. Then again... that's not technically _our_ problem; the sync feedback would fail to come back either way.

rasifr: Correct. However, we only need to determine whether a synchronous standby has been requested, not whether its actually reachable. I believe this check can be replaced by synchronous_standby_names GUC, which can determine if synchronous replication has been configured.

ibrarahmad:
No, we need to ensure that the replica is reachable too, so we need
that status.

On Wed, May 28, 2025 at 8:02 PM Asif Rehman ***@***.***> wrote:
>
> @rasifr commented on this pull request.
>
> ________________________________
>
> In spock_apply.c:
>
> > @@ -994,6 +1018,9 @@ handle_commit(StringInfo s)
>
>   CommitTransactionCommand();
>
> + if (WalSndCtl->sync_standbys_status & SYNC_STANDBY_DEFINED)
>
> Correct. However, we only need to determine whether a synchronous standby has been requested, not whether its actually reachable. I believe this check can be replaced by synchronous_standby_names GUC, which can determine if synchronous replication has been configured.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you authored the thread.Message ID: ***@***.***>



-- 
Ibrar Ahmed

spock_apply.c
ibrarahmad: Fixed

spock_apply.c
ibrarahmad: No, we only need to ensure that the replica is reachable, so we need only that status.

ibrarahmad: @rasifr Secondly, this change has been backported to PostgreSQL-13, so there is no difference between the new and old versions of PostgreSQL.

rasifr (approved):
LGTM

PR #56, opened 2025-06-04
Fix progress entry creation timing and correct remote_node_id usage i…

rasifr:
…n sync update

- Moved create_progress_entry() earlier in spock_create_subscription() to ensure the progress entry exists before sync status creation.

- Fixed SQL query in adjust_progress_info() to use the correct remote_node_id, ensuring the update targets the appropriate row for the newly created subscription.


ibrarahmad (approved):
LGTM

PR #66, opened 2025-06-18
Always log apply conflicts; remove unused before-trigger flag

rasifr:
Changed spock_report_conflict() to always log conflict information, regardless of which tuple (local or remote) wins the resolution. This improves observability and debugging.

Improved delete conflict handling by logging instead of erroring when the target row is missing. Also removed the unused 'has_before_triggers' flag from conflict logging APIs.


ibrarahmad (approved):
LGTM

PR #70, opened 2025-06-25
Fix/SPOC-57: Prevent PK/RI-less INSERTs from triggering upsert logic

rasifr:
Updated spock_apply_heap to skip replica identity lookups for INSERTs without primary keys or replica identity, avoiding unnecessary sequential scans and preventing inserts from incorrectly entering upsert/conflict-handling paths.


ibrarahmad (approved):
LGTM

PR #74, opened 2025-06-26
Fix/spoc 60: Missing default values, Along with enabled partials unit tests

rasifr:


mason-sharp (approved):
SPOC-60 got pulled into this

PR #78, opened 2025-07-02
[SPOC-53] Improve spockctrl behavior and add-node workflow for PG upgrades.

ibrarahmad:
[SPOC-53] Improve spockctrl behavior and add-node workflow for PG upgrades.

- Display help message even if spockctrl.json is not specified.
- Fix minor issue in add-node workflow for PostgreSQL major version upgrade.


AntTheLimey (approved):
LGTM.

rasifr (approved):

PR #81, opened 2025-07-08
Fix spockbench tests

pct960:


mason-sharp:
LGTM, but I would like @rasifr to take a look as well.

I am also ok with keeping 300 seconds instead of 120. What would make me feel better is if we modified spockbench to be able to pass in the number of accounts per branch, so that we can use something like 100 - 200 accounts per branch to generate more conflicts. Yes, it deviates from the 100k default, but for our purposes here as part of our CI we want to make sure we have conflicts. I can add a ticket to the backlog.

wieck:
> LGTM, but I would like @rasifr to take a look as well.
> 
> I am also ok with keeping 300 seconds instead of 120. What would make me feel better is if we modified spockbench to be able to pass in the number of accounts per branch, so that we can use something like 100 - 200 accounts per branch to generate more conflicts. Yes, it deviates from the 100k default, but for our purposes here as part of our CI we want to make sure we have conflicts. I can add a ticket to the backlog.

You can increase the rate of conflicts by changing the transaction mix. For example using the transaction that picks a random teller (which is not used by default).

pct960:
The `spock.resolutions` table gets filled up pretty quickly as the test progresses. So, I'd say we definitely have a pretty contentious workload. And, as for the test duration, @mason-sharp, I'll change it back to five minutes.

rasifr (approved):
LGTM. I would prefer to keep CI runs on the lower end -- 5 minutes is more than enough. Anything longer should be handled by a separate periodic workflow that runs independently of PRs.

PR #82, opened 2025-07-09
[SPOC-83] Refactor test container for running regression tests

mason-sharp:
Previously Postgres and Spock were compiled in entrypoint.sh as part of running spockbench. This is now moved to building these in the container, allowing us to use the same container to run regression tests as a separate step that can also be more easily viewed in the github workflow output.

Screenshot 2025-07-08 at 2 02 31 PM


tests/Dockerfile.el9
```diff
@@ -19,9 +23,64 @@ RUN dnf install --allowerasing --enablerepo=crb -y $(cat /home/pgedge/lib-list.t
...
 
-COPY entrypoint.sh run-tests.sh /home/pgedge
 
-RUN sudo chmod +x /home/pgedge/entrypoint.sh /home/pgedge/run-tests.sh
+#-----------------------------------------
+USER pgedge
+
+WORKDIR /home/pgedge
+RUN echo "Cloning Spock at $GITHUB_REF from $REPO_URL"
+RUN git clone ${REPO_URL}
```
pct960, line 32: The repo is already checked out by the action. For the next update, we could perhaps simply copy the files while building the image.

PR #83, opened 2025-07-09
Fix primary_key regression test to align with Spock 5.0 behavior

rasifr:
Updated the test case to reflect current behavior where workers no longer quit on replication errors. With Spock 5.0, subscriptions are now disabled when exception_behaviour = 'SUB_DISABLE'.

Also added a helper function to the init script to simplify setup of test scenarios.


mason-sharp:
The codacy complaints are for the regression test sql files that run fine and can be ignored. Will look into configuring to ignore those.

PR #88, opened 2025-07-11
Trigger tests on PRs from forked repos

mason-sharp:
Make sure that PRs that come from forked repos also trigger the testing github action.


mason-sharp:
Closing since more work is needed to handle both same-repo branches and forked-repo branches

PR #98, opened 2025-07-22
Fixed triggers regression test

rasifr:


spock--5.0.0.sql
```diff
@@ -318,7 +318,7 @@ CREATE TABLE spock.queue (
 
 CREATE FUNCTION spock.replicate_ddl(command text,
 									replication_sets text[] DEFAULT '{ddl_sql}',
-									search_path text DEFAULT '',
+									search_path text DEFAULT current_setting('search_path'),
```
mason-sharp, line 321: Should this go in a `spock--5.0.0--5.0.1.sql` file as a `CREATE OR REPACE FUNCTION` instead (and we first get in a version change PR)?

mason-sharp (changes requested):

expected/att_list.out
```diff
@@ -299,7 +299,7 @@ SELECT * FROM basic_dml ORDER BY id;
 SELECT spock.replicate_ddl($$
 	DROP TABLE public.basic_dml CASCADE;
 $$);
-NOTICE:  drop cascades to table public.basic_dml membership in replication set default
+NOTICE:  drop cascades to table basic_dml membership in replication set default
```
mason-sharp, line 302: IIUC, if search_path is set and matches, it will leave off the schema namespace, but if it is in a different schema it will include it?

rasifr: Thats true. So far the current tests only make use of public schema and matches to search_path. But later if we add a test with different schema, the notices will have schema qualification.

spock--5.0.0.sql
rasifr: Right now we don’t have an upgrade script from the v5_STABLE branch to the main branch version. Once the version in main is bumped (e.g., to 5.1.0 or 6.0.0), and the corresponding upgrade script is added, we’ll include this change there.

PR #122, opened 2025-08-11
[SPOC-144] Reduce memory usage

mason-sharp:
When running a smaller, modified version of the huge 100k table test that only uses 15k tables, peak memory usage was 19.8 GB. With the changes in this PR, it is reduced to 647 MB. The memory context MessageContext is used for the entire duration of a transaction, so large transactions will use a lot of memory unless we free memory along the way. A new memory context ApplyOperationContext is introduced and used for handling INSERTs.

Creating similar tests for UPDATE and DELETE did not show the same memory usage problem as INSERT thankfully, so the changes here are limited to the INSERT case.

### DIFF
--- a/spock_apply.c
+++ b/spock_apply.c
@@ -257,6 +257,12 @@ dlist_head sync_replica_lsn = DLIST_STATIC_INIT(sync_replica_lsn);
 static XLogRecPtr skip_xact_finish_lsn = InvalidXLogRecPtr;
 #define is_skipping_changes() (unlikely(!XLogRecPtrIsInvalid(skip_xact_finish_lsn)))
 
+/*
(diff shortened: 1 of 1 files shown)


spock_apply.c
```diff
@@ -1473,7 +1486,9 @@ handle_insert(StringInfo s)
 	}
 	else
 	{
+		oldcontext = MemoryContextSwitchTo(ApplyOperationContext);
```
ibrarahmad: @mason-sharp I checked that at the beginning of the function, we assign a value to oldcontext, and then we reassign it later. Is this correct during the final cleanup?

mason-sharp: Good catch, I removed the reassignment.

PR #151, opened 2025-09-10
Spock resource manager

rasifr:
**Spock: Custom WAL RMGR + shmem apply-progress; clean-shutdown snapshot**

- Add Spock-specific WAL resource manager and log apply-progress records.
- Keep progress in shmem (per-group entry) and Removed the catalog table spock.progress
- On clean shutdown, snapshot progress to $PGDATA/spock/resource.dat; reload on startup.

**Design**

- WAL -> shmem pipeline:
  - Apply emits a compact progress record via the Spock RMGR.
  - Shmem group entry is the live state workers read/update.
  - Crash case - Redo rebuilds shmem from WAL. Clean restart - Load from resource.dat.

- Grouped progress:
  - Each subscription attaches to a group entry tracking last remote commit ts/LSNs, etc.
  - Workers reference the entry directly (attach/detach); no per-commit catalog DML.

- Clean-shutdown snapshot:
  - on_shmem_exit walks the group table and writes a versioned snapshot.
  - write-temp + fsync + durable_rename to avoid torn files.
  - Startup validates version/system_identifier; on mismatch/corruption, ignore snapshot and continue.

**Rationale**

Catalog writes in the apply hot path were heavyweight and would cause
    locking and bloating issues, under heavy loads. By WAL-logging
    progress and rebuilding in-memory state during redo:
      * Apply path is lighter weight (no catalog DML per commit).
      * Progress tracking is isolated from user activity and schema changes.

### DIFF
--- a/sql/spock--6.0.0-devel.sql
+++ b/sql/spock--6.0.0-devel.sql
@@ -95,16 +95,24 @@ CREATE TABLE spock.exception_status_detail (
 		REFERENCES spock.exception_status
 ) WITH (user_catalog_table=true);
 
-CREATE TABLE spock.progress (
-	node_id oid NOT NULL,
-	remote_node_id oid NOT NULL,
-	remote_commit_ts timestamptz NOT NULL,
-	remote_lsn pg_lsn NOT NULL,
-	remote_insert_lsn pg_lsn NOT NULL,
-	last_updated_ts timestamptz NOT NULL,
-	updated_by_decode bool NOT NULL,
-	PRIMARY KEY(node_id, remote_node_id)
-) WITH (fillfactor=50);
+CREATE FUNCTION spock.apply_group_progress (
+	OUT dbid oid,
+	OUT node_id oid,
+	OUT remote_node_id oid,
+	OUT remote_commit_ts timestamptz,
+	OUT prev_remote_ts timestamptz,
+	OUT remote_commit_lsn pg_lsn,
+	OUT remote_insert_lsn pg_lsn,
+	OUT last_updated_ts timestamptz,
+	OUT updated_by_decode bool
+) RETURNS SETOF record
+LANGUAGE c AS 'MODULE_PATHNAME', 'get_apply_group_progress';
+
+CREATE VIEW spock.progress AS
+	SELECT node_id, remote_node_id, remote_commit_ts,
+		   remote_commit_lsn AS remote_lsn, remote_insert_lsn,
+		   last_updated_ts, updated_by_decode
+	FROM spock.apply_group_progress();
 
 CREATE FUNCTION spock.node_create(node_name name, dsn text,
     location text DEFAULT NULL, country text DEFAULT NULL,
--- a/src/spock.c
+++ b/src/spock.c
@@ -55,6 +55,7 @@
 #include "spock_executor.h"
 #include "spock_node.h"
 #include "spock_conflict.h"
+#include "spock_rmgr.h"
 #include "spock_worker.h"
 #include "spock_output_plugin.h"
 #include "spock_exception_handler.h"
@@ -917,7 +918,7 @@ _PG_init(void)
 	if (!process_shared_preload_libraries_in_progress)
 		elog(ERROR, "spock is not in shared_preload_libraries");
 
-	DefineCustomEnumVariable("spock.conflict_resolution",
+    DefineCustomEnumVariable("spock.conflict_resolution",
 							 gettext_noop("Sets method used for conflict resolution for resolvable conflicts."),
 							 NULL,
 							 &spock_conflict_resolver,
@@ -1155,6 +1156,9 @@ _PG_init(void)
 	if (IsBinaryUpgrade)
 		return;
 
+	/* Spock resource manager */
+	spock_rmgr_init();
+
 	/* Init workers. */
 	spock_worker_shmem_init();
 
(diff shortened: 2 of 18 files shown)


src/spock.c
```diff
@@ -801,7 +802,7 @@ _PG_init(void)
 	if (!process_shared_preload_libraries_in_progress)
 		elog(ERROR, "spock is not in shared_preload_libraries");
 
-	DefineCustomEnumVariable("spock.conflict_resolution",
```
danolivo, line 920: What is the difference here between old and new versions?

src/spock_group.c
```diff
@@ -0,0 +1,282 @@
...
+	int			napply_groups;
+	Size		size;
+
+#if PG_VERSION_NUM >= 150000
+	if (prev_shmem_request_hook != NULL)
+		prev_shmem_request_hook();
+#endif
+
+	/*
+	 * This is cludge for Windows (Postgres des not define the GUC variable as
```
danolivo: Just a typo

src/spock_group.c
```diff
@@ -0,0 +1,282 @@
...
+								   HASH_SHARED_MEM | HASH_PARTITION |
+								   HASH_FIXED_SIZE);
+
+	if (!SpockGroupHash)
+		elog(ERROR, "spock_group_shmem_startup: failed to init group map");
+
+	LWLockRelease(AddinShmemInitLock);
+}
+
+static inline SpockGroupKey
```
danolivo, line 173: It seems Oid is a 4-byte element. I suppose different platforms will organise this structure with different paddings between fields, no? In this case does we have a guarantee that during allocation paddings will not filled with a waste and memcmp() will not fail on comparison of two such structs with similar values in these fields? May be better to use memset(0) to be sure?

src/spock_group.c
```diff
@@ -0,0 +1,282 @@
...
+{
+	SpockGroupKey key = make_key(dbid, node_id, remote_node_id);
+	SpockGroupEntry *e;
+	bool		found;
+
+	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
+	if (!found)
+	{
+		/* initialize new entry */
+		e->key = key;
```
danolivo: Unnecessary field, isn't it?

src/spock_group.c
```diff
@@ -0,0 +1,282 @@
...
+
+	if (!sap)
+		return false;
+
+	key = make_key(sap->key.dbid, sap->key.node_id, sap->key.remote_node_id);
+	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
+
+	if (!found)					/* New Entry */
+	{
+		e->key = key;
```
danolivo: Unnecessary field

src/spock_worker.c
```diff
@@ -310,15 +310,28 @@ wait_for_worker_startup(SpockWorker *worker,
...
 /*
  * Cleanup function.
- *
- * Called on process exit.
+ */
+
+/*
+ * Called on shmem exit.
+ */
+static void
```
danolivo, line 316: This function will be called at each backend and, if we shut down 100 backends at the same time it may cause overwriting one file multiple times ...
Also, on_shmem_exit seems too late to write files. Maybe to do it a little earlier? Or just let to write this file by the postmaster only? If we do it only on a standard shutdown it may work properly. Also, there are some check should be to avoid writing this file on fast shutdown.

src/spock_common.c
```diff
@@ -519,3 +521,96 @@ spock_build_replindex_scan_key(ScanKey skey, Relation rel, Relation idxrel,
...
+			if (errno == EINTR)
+				continue; /* retry */
+
+			/* One consolidated error path: EOF or hard error */
+			if (n < 0)
+				ereport(ERROR,
+						(errcode(ERRCODE_DATA_CORRUPTED),
+						errmsg("could not read file \"%s\": %m", fname)));
+
+			ereport(ERROR,
```
mason-sharp, line 559: Maybe this has something to do with the codacy complaint. Can this report an error (false positive) if after the first time through the loop 0 bytes are read?

src/spock_worker.c
rasifr: Thats a good catch. Yes its meant to write the file under postmaster only; not for all backends.

src/spock_group.c
rasifr: Thats true. Will memset(0) it.

src/spock.c
rasifr: Nothing. Likely formatting got fixed by pgindent.

src/spock_common.c
rasifr: The intended behaviour for this function is to error out in case the required bytes cannot be read. whether its because of EOF or any negative return. If the desired bytes have been read, the loop should terminate early.

sql/spock--6.0.0-devel.sql
```diff
@@ -95,16 +95,24 @@ CREATE TABLE spock.exception_status_detail (
 		REFERENCES spock.exception_status
 ) WITH (user_catalog_table=true);
 
-CREATE TABLE spock.progress (
```
mason-sharp, line 98: We should include an upgrade script as well.

mason-sharp (changes requested):

PR #152, opened 2025-09-10
Fix minor issues in core patches to avoid warnings (PG18 only).

danolivo:
This is mostly as a part of onboarding to be sure I understand the whole development cycle. But it also annoying to see it each compile time.

These warning is shown only in Postgres 18, caused by the 962da90. Also, let the spockctrl choose the same compiler as the Postgres core has used.
XXX: Should we use INT64_FORMAT for a timestamp instead?

Also here I propose to use for spockctrl the same compiler as was used for the Postgres core.


danolivo:
> > XXX: Should we use INT64_FORMAT for a timestamp instead?
> > Yes definitely. Otherwise, it will generate warning when compiling with older versions.
Ok, done. Not sure how this slight inconsistency in the output format might have influenced older versions - so don't back-patch it yet.

mason-sharp (approved):

PR #153, opened 2025-09-15
TAP test helpers

mason-sharp:
Some new functions to make writing and reading the code for TAP tests a little easier.

`scalar_query` take the node number (starting from 1) and a statement to run, expecting one single value.

`psql_or_die` is like `system_or_die`, but runs a psql command.

`001_basic.pl` is changed to show using these. The other existing test files have not been modified yet.

### DIFF
--- a/tests/tap/t/001_basic.pl
+++ b/tests/tap/t/001_basic.pl
@@ -2,7 +2,7 @@ use strict;
 use warnings;
 use Test::More tests => 29;
 use lib '.';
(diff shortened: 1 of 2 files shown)


rasifr (commented):
Should `psql_or_bail` and `scalar_query` accept the port directly as argument?

mason-sharp:
> Should `psql_or_bail` and `scalar_query` accept the port directly as argument?

It passes in the node number (starting from 1), from that we know the port that it is using.

ibrarahmad (approved):
LGTM

PR #158, opened 2025-09-17
SPOC-51: obfuscate DSN password in error logs

danolivo:
Includes:

1. Wipe out password from error message
2. Lower log level for a couple of core messages


src/spock.c
```diff
@@ -789,6 +791,119 @@ spock_temp_directory_assing_hook(const char *newval, void *extra)
...
+		strpos = pg_strcasestr(debug_query_string, substr);
+		if (strpos != NULL)
+		{
+			strpos += len;
+			replace_symbols(strpos);
+		}
+	}
+
+	/*
+	 * Filter messages that previously needed core patch 'pg18-020-LOG-to-DEBUG1'
```
mason-sharp, line 879: Good idea here!

If no longer needed, please go ahead and include deleting the patch file in this PR for all Postgres versions.

Perhaps add clarifying comments that with Spock we always want these cases logged as LOG, instead of only with DEBUG1.

src/spock.c
```diff
@@ -789,6 +791,119 @@ spock_temp_directory_assing_hook(const char *newval, void *extra)
...
+}
+
+/* Replace each following symbol with 'X' until the EOL */
+static void
+replace_symbols(char *strpos)
+{
+	if (strpos == NULL)
+		return;
+
+	while (*strpos != '\0')
```
mason-sharp, line 826: Just a comment that I was originally thinking we could end when encountering a " " or ";", but your implementation seems safer.

src/spock.c
```diff
@@ -789,6 +791,120 @@ spock_temp_directory_assing_hook(const char *newval, void *extra)
...
+		prev_emit_log_hook(edata);
+
+	if (!edata->output_to_client && !edata->output_to_server)
+		/* Previous hook already done this job. */
+		return;
+
+	if (edata->elevel == ERROR)
+	{
+		char	   *strpos;
+		const int	len = strlen(PSWD_KEYWORD);
```
mason-sharp, line 857: I wonder if codacy would not complain if we use `sizeof` here since it is compile-time on a string literal.

PR #165, opened 2025-09-19
SPOC-223, SPOC-224, SPOC-225, SPOC-233: Update ZODAN flow, add remove-node SQL and Python scripts.

ibrarahmad:
SPOC-223: Update ZODAN flow, add remove-node SQL and Python scripts.
 
 - Check subscription state after enabling the disabled subscription.
 - Improve phase logging and progress output with clear, terse lines.
 - Default timeout is 1200s. Document param as timeout_ms.

SPOC-224: Remove-node tools

  - Add zodrn.sql: ordered removal of repsets, subscriptions, slots, and node registry. Use guarded transactions and status checks.

  - Add zodrn.py: Python peer using direct connections, not dblink. Mirror zodrn.sql steps with retries and exit codes.

SPOC-225: New phase to verify syncing process

  - Fire sync_event right after creation of disabled replication slots.
  - Enable the subscription, then call wait_for_sync_event to gate progress.

Logging and summary

   - Prefix logs with zodan. Include node ids, LSNs, and step timings.
   - Provide a final summary: objects removed, slots dropped, node id freed.


mason-sharp:
I get syntax errors trying to source both zodan.sql and zodrn.sql to test them out. I only tried against PG 18.

PR #167, opened 2025-09-19
SPOC-208: Treat same origin updates as non-conflicts

mason-sharp:
As of Spock 5, we were treating updates to rows that originated from that same origin as a conflict, as well as rows that get updated that were inserted in the same transaction. We no longer log these in the `spock.resolutions` table.

Added additional tests to `sql/tuple_origin.sql`

### DIFF
--- a/src/spock_conflict.c
+++ b/src/spock_conflict.c
@@ -691,6 +691,23 @@ spock_report_conflict(SpockConflictType conflict_type,
 	const char *idxname = "(unknown)";
 	const char *qualrelname;
 
+
+	/* Ignore update-update conflict for same origin */
+	if (conflict_type == CONFLICT_UPDATE_UPDATE)
+	{
+		/*
+		 * If updating a row that came from the same origin,
+		 * do not report it as a conflict
+		 */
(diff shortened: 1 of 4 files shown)


src/spock_conflict.c
```diff
@@ -691,6 +691,23 @@ spock_report_conflict(SpockConflictType conflict_type,
...
 	const char *qualrelname;
 
+
+	/* Ignore update-update conflict for same origin */
+	if (conflict_type == CONFLICT_UPDATE_UPDATE)
+	{
+		/*
+		 * If updating a row that came from the same origin,
+		 * do not report it as a conflict
+		 */
```
rasifr, line 701: Whats about UPDATE-DELETE conflict from the same origin?

mason-sharp: IIUC- if we get an UPDATE for a row that has been deleted, we currently don't track tombstone records, so we would not know where the previous deleted row came from. If we did track it and the deletion came from the same origin before the UPDATE, it would imply we missed an insert somewhere from another node.

PR #168, opened 2025-09-22
SPOC-102 - INSERT into an absent table shouldn't halt the apply worker

danolivo:
TODO: UPDATE and DELETE should be processed the same way if we agree it is a right way.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -1338,7 +1338,7 @@ handle_relation(StringInfo s)
 }
 
 static void
-log_insert_exception(bool failed, ErrorData *edata, SpockRelation *rel,
+log_insert_exception(bool failed, char *errmsg, SpockRelation *rel,
 					 SpockTupleData *oldtup, SpockTupleData *newtup,
 					 const char *action_name)
 {
@@ -1374,7 +1374,7 @@ log_insert_exception(bool failed, ErrorData *edata, SpockRelation *rel,
 							   rel, localtup, oldtup, newtup,
 							   NULL, NULL,
 							   action_name,
-							   (failed && edata) ? edata->message : NULL);
+							   (failed) ? errmsg : NULL);
 }
 
 static void
@@ -1382,7 +1382,7 @@ handle_insert(StringInfo s)
 {
 	SpockTupleData newtup;
 	SpockRelation *rel;
-	ErrorData  *edata;
+	ErrorData  *edata = NULL;
 	MemoryContext	oldcontext;
 	bool		started_tx;
 	bool		failed = false;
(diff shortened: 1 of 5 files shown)


src/spock_apply.c
```diff
@@ -1397,11 +1397,28 @@ handle_insert(StringInfo s)
...
+		 */
+		xact_had_exception = true;
+		exception_command_counter++;
+		log_insert_exception(true, "Spock can't find relation", NULL,
+							 NULL, NULL, "INSERT");
+		end_replication_step();
+		MemoryContextSwitchTo(oldcontext);
+		MemoryContextReset(ApplyOperationContext);
+		return;
+	}
```
rasifr, line 1417: I think this may also affect UPDATE and DELETE statements; maybe we should account for those too?

src/spock_apply.c
```diff
@@ -3136,12 +3205,13 @@ apply_work(PGconn *streamConn)
 			 */
 			if (MyApplyWorker->use_try_block)
 			{
-				elog(LOG, "SPOCK: caught exception while use_try_block=true");
+				elog(LOG, "SPOCK: caught exception while use_try_block=true. The exception was:\n'%s'",
```
mason-sharp, line 3208: 👍

tests/regress/expected/replication_set.out
```diff
@@ -231,3 +231,408 @@ SELECT * FROM spock.replication_set;
...
+SELECT * FROM spoc_102l ORDER BY x; -- ERROR, does not exist yet
+ERROR:  relation "spoc_102l" does not exist at character 15
+-- Now, fix the issue with absent table
+BEGIN;
+SELECT spock.repair_mode(true) \gset
+CREATE TABLE spoc_102l (x integer PRIMARY KEY);
+END;
+-- Check that replication works
+INSERT INTO spoc_102l VALUES (4);
+-- XXX: Why we don't synchronize the state of the table and don't see rows
```
mason-sharp, line 307: Are they the rows in the `spock.exception_log` table when it could not find the relation?

rasifr, line 307: > -- XXX: Why we don't synchronize the state of the table and don't see rows

@danolivo this is due to the transaction discard mode. Which rolls back the bad transactions. If the transaction mode were set to `SUB_DISABLE`, the subscription would be suspended. After fixing the issue and the subscription is enabled, the earlier rows that were added by publisher would be pulled again and should be present in the table.

In transdiscard or discard mode, these rows should be present in the `exception_log` table.

mason-sharp: Meaning, later below in the output I believe those are expected entries in the `spock.exception_log` table.

PR #169, opened 2025-09-22
Doc updates for Spock 5.0.2

susan-pgedge:
This is for SPOC-234; add a schema qualifier to the front of the call for Zodan, updates the release notes, and will add changes for Zodan.

### DIFF
--- a/docs/zodan.md
+++ b/docs/zodan.md
@@ -170,7 +170,7 @@ SELECT spock.sub_create(
     provider_dsn := 'host=127.0.0.1 dbname=inventory port=5432 user=alice password=1safepassword',
     replication_sets := ARRAY['default', 'default_insert_only', 'ddl_sql'],
     synchronize_structure := true,
-    skip_schema := ARRAY['management-only'],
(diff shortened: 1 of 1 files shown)


docs/zodan.md
```diff
@@ -162,6 +179,15 @@ SELECT spock.sub_create(
 );
 ```
 
+Optionally, include the `skip_schema` parameter and a comma-delimited array of schemas that you would like to omit when synchronizing the database structure.  The parameter can be useful in cases where the source machine contains information or extensions that you do not wish to propagate to other nodes on your network.
+
+The `skip_schema` parameter is only enforced when `synchronize_structure` is set to `true`.  For example, the following code snippet omits a schema named `management-only` from the subscription:
+
+```sql
+synchronize_structure = true,
+skip_schema := ARRAY['mgmt-only'],
```
mason-sharp, line 188: Does not match `management-only`.

mason-sharp (changes requested):

docs/zodan.md
susan-pgedge: Updated...

PR #171, opened 2025-09-23
Remove snowflake-related functions and doc from the Spock.

danolivo:
It seems obsolete code that doesn't even work correctly.


sql/spock--6.0.0-devel.sql
```diff
@@ -518,201 +518,6 @@ CREATE FUNCTION spock.terminate_active_transactions() RETURNS bool
  AS 'MODULE_PATHNAME', 'spockro_terminate_active_transactions'
  LANGUAGE C STRICT;
 
--- ----------------------------------------------------------------------
--- We check the PostgreSQL major version number in case a future
--- catalog change forces us to provide different functions for
--- different versions.
--- ----------------------------------------------------------------------
-DO $version_dependent$
```
mason-sharp, line 534: @rasifr @ibrarahmad do you know the history of this snippet?

sql/spock--6.0.0-devel.sql
```diff
@@ -518,201 +518,6 @@ CREATE FUNCTION spock.terminate_active_transactions() RETURNS bool
...
-	CASE
-		WHEN pgmajor IN (15, 16, 17, 18) THEN
-
--- ----------------------------------------------------------------------
--- convert_column_to_int8()
---
---	Change the data type of a column to int8 and recursively alter
---	all columns that reference this one through foreign key constraints.
--- ----------------------------------------------------------------------
-CREATE OR REPLACE FUNCTION spock.convert_column_to_int8(p_rel regclass, p_attnum smallint)
```
mason-sharp, line 549: Once PR #181 is approved, I think we could DROP these in spock--5.0.4--6.0.0-devel.sql as well.

mason-sharp (changes requested):

sql/spock--6.0.0-devel.sql
rasifr: IIRC, this might be because PG 14 was still supported by the Spock 3.x extension at the time Snowflake support was introduced. However, snowflake was not supported for PG 14 -- hence the need for the version check.

sql/spock--6.0.0-devel.sql
mason-sharp: PR 181 was merged.

danolivo: > Once PR #181 is approved, I think we could DROP these in spock--5.0.4--6.0.0-devel.sql as well.

Done. And rebased on the fresh main

PR #192, opened 2025-10-01
Generate subscription ID

danolivo:
Emulate behaviour of Oid type: don't allow negative and zero values here. In principle, the sub_id type should be changed to int32 or int64. However, it would immediately cause numerous changes throughout the code. And what's worse, it will cause the Spock UI change, which may need changes to the CLI and other components...
Therefore, simply change the value generator from hash_any to a safer and more stable sequence.

Add sequence 'sub_id_generator' in the UI of Spock.

The output of some tests was changed because of uncovered sub_id values. It would be better to hide this value as unstable by definition, but let it go for now to simplify the commit and keep in mind the good enough stability of the sequence.

### DIFF
--- a/sql/spock--5.0.4--6.0.0-devel.sql
+++ b/sql/spock--5.0.4--6.0.0-devel.sql
@@ -45,3 +45,7 @@ CREATE VIEW spock.lag_tracker AS
 	LEFT JOIN spock.node origin ON sub.sub_origin = origin.node_id
 	LEFT JOIN spock.node n ON n.node_id = p.node_id
 	GROUP BY origin.node_name, n.node_name;
+
+-- Source for sub_id values.
+CREATE SEQUENCE spock.sub_id_generator AS integer MINVALUE 1 CYCLE START WITH 1
+OWNED BY spock.subscription.sub_id;


tests/regress/expected/init.out
```diff
@@ -46,10 +46,10 @@ BEGIN
...
 \dx spock
-                 List of installed extensions
- Name  |   Version   | Schema |          Description           
--------+-------------+--------+--------------------------------
- spock | 6.0.0-devel | spock  | PostgreSQL Logical Replication
+                          List of installed extensions
+ Name  |   Version   | Default version | Schema |          Description           
+-------+-------------+-----------------+--------+--------------------------------
+ spock | 6.0.0-devel | 6.0.0-devel     | spock  | PostgreSQL Logical Replication
 (1 row)
```
ibrarahmad, line 59: Can we avoid having a version in the output? Instead of using \dx, we query the catalog and do not select a version. In that case, we can avoid changing the version on every release.

src/spock_node.c
```diff
@@ -765,6 +766,41 @@ get_node_interface_by_name(Oid nodeid, const char *name, bool missing_ok)
 	return nodeif;
 }
 
+/*
+ * Generate subscription ID.
```
ibrarahmad, line 770: I like that idea in general, but we need to test the whole system, specifically ZODAN.

tests/regress/expected/init.out
danolivo: We may query the pg_extension directly and manage output according to the purpose.

src/spock_node.c
danolivo: Ok. Hence, no so close to commit ;).

tests/regress/expected/init.out
danolivo: Done!

sql/spock--6.0.0-devel.sql
```diff
@@ -37,6 +37,8 @@ CREATE TABLE spock.subscription (
 	sub_skip_lsn pg_lsn NOT NULL DEFAULT '0/0',
 	sub_skip_schema text[]
 );
+-- Source for sub_id values.
+CREATE SEQUENCE spock.sub_id_generator AS integer MINVALUE 1 CYCLE START WITH 1 OWNED BY spock.subscription.sub_id;
```
mason-sharp, line 41: We should also include it in an upgrade sql file.

danolivo: Done.

PR #193, opened 2025-10-01
fix: guard NULL from spock_relation_open() and fail with clear error

rasifr:
When a table is added to a repset and the subscription begins data sync, Spock monitors for the table's presence and starts copying once it appears.

Now spock_relation_open() is changed to return NULL (instead of throwing error) for not-yet-present relations. The callers in sync and truncate paths could continue using the invalid rel, leading to invalid memory access and causing segmentation fault.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -1803,6 +1803,11 @@ handle_truncate(StringInfo s)
 		Oid			relid = lfirst_oid(lc);
 
 		rel = spock_relation_open(relid, lockmode);
+		if (rel == NULL)
+			ereport(ERROR,
+					(errcode(ERRCODE_UNDEFINED_TABLE),
+					 errmsg("Spock can't find relation with oid %u", relid)));
+
 		errcallback_arg.rel = rel;
 
 		/* If in list of relations which are being synchronized, skip. */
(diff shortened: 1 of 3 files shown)


src/spock_apply.c
```diff
@@ -1803,6 +1803,11 @@ handle_truncate(StringInfo s)
 		Oid			relid = lfirst_oid(lc);
 
 		rel = spock_relation_open(relid, lockmode);
+		if (rel == NULL)
```
danolivo, line 1806: It would make sense to use the same exception logic like in the case of INSERT or UPDATE. But I understand why this simplification may be used here.

tests/tap/t/004_non_default_repset.pl
```diff
@@ -81,12 +81,12 @@ BEGIN
 system_or_bail "$pg_bin/psql", '-p', $node_ports->[0], '-d', $dbname, '-c', "SELECT spock.repset_remove_table('default', 'basic_dml1')";
 
 # Check if table is already in delay replication set, if not add it
-my $table_in_delay = `$pg_bin/psql -p $node_ports->[0] -d $dbname -t -c "SELECT EXISTS (SELECT 1 FROM spock.repset_table WHERE set_name = 'delay' AND set_relname = 'basic_dml1')"`;
```
danolivo, line 84: Not sure: may be TRUNCATE on an absent table deserves regression test? Is this possible to invent a test on table_sync too?

danolivo (approved):

PR #194, opened 2025-10-02
SPOC-193: autoddl: guard extension-script subcommands using creating_extension

rasifr:
During upgrades, the extension script runs a bunch of internal subcommands with creating_extension = true. However, Our autoddl was using ProcessUtilityContext to skip subcommands; which sometimes let these slip through, causing unintended replication.

### DIFF
--- a/src/spock_executor.c
+++ b/src/spock_executor.c
@@ -356,6 +356,10 @@ autoddl_can_proceed(ProcessUtilityContext context, NodeTag toplevel_stmt,
 	if (context == PROCESS_UTILITY_TOPLEVEL)
 		return true;
 
+	/* Guard against CREATE EXTENSION subcommands */
+	if (creating_extension)
+		return false;
+
 	/*
 	 * Indicates a portion of a query. These statements should be handled by the
 	 * corresponding top-level query.
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ REGRESS = preseed infofuncs init_fail init preseed_check basic conflict_secondar
 		  toasted replication_set matview bidirectional primary_key \
 		  interfaces foreign_key copy sequence triggers parallel functions row_filter \
 		  row_filter_sampling att_list column_filter apply_delay \
-		  extended node_origin_cascade multiple_upstreams tuple_origin \
+		  extended node_origin_cascade multiple_upstreams tuple_origin autoddl \
 		  drop
 
 # The following test cases are disabled while developing.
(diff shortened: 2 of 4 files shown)


tests/tap/t/009_auto_ddl.pl
```diff
@@ -0,0 +1,82 @@
...
+ok(wait_until(30, sub {
+  scalar_query(2, "SELECT status FROM spock.sub_show_status('test_sub_n2')") eq 'replicating'
+}), 'subscription is replicating');
+
+# Wait for sync event to be processed on subscriber (node 2)
+psql_or_bail(2, "CALL spock.wait_for_sync_event(true, 'n1', $syncevent)");
+
+diag(psql_or_bail(2, "SELECT * FROM spock.tables"));
+# Check schema, table and function appear on subscriber (node 2)
+ok(scalar_query(2, "SELECT count(*) FROM spock.tables where nspname = 'hollywood' AND set_name IS NOT NULL") eq '2',
```
mason-sharp: It looks like this test is failing in that it does not exist.

I wonder if it created the schema since it would be a top level command.

I took a look at PostgreSQL contrib modules, none of them seem to create a custom schema, but if our test container is not installing contrib modules, perhaps we could compile and install one on the source, run CREATE EXTENSION, then verify some custom function it creates does not appear on the destination.

src/spock_executor.c
```diff
@@ -356,6 +356,10 @@ autoddl_can_proceed(ProcessUtilityContext context, NodeTag toplevel_stmt,
 	if (context == PROCESS_UTILITY_TOPLEVEL)
 		return true;
 
+	/* Guard against CREATE EXTENSION subcommands */
+	if (creating_extension)
```
danolivo, line 360: I see here two issues:
1. Why to skip replication of ANY extension?
2. What about ALTER/DROP command?
I suggest we need to go a little more smart way, may be cancelling only the Spock manipulations - we already have ready query plan tree at this point ...

rasifr: We are skipping the objects created by extension scripts, but not the extension itself. What we need is to replicate the user-issued command, i.e: CREATE EXTENSION foo; — without replicating the underlying objects that the extension foo creates.

And yes, this code segment needs a refactoring which is coming up next.

mason-sharp: Adding that `creating_extension` is also set to `true` in Postgres when executing an `ALTER EXTENSION`.  During `DROP EXTENSION`, dependencies are used to remove objects.

mason-sharp (approved):

PR #212, opened 2025-10-09
Increase stability of regression tests (spoc-197).

danolivo:
There are a few tests that rely on the time period required to synchronise a subscription or deliver replicated data to a subscriber. Such time dependency causes tests' instability, which can be easily checked by inserting something like 'pg_usleep(10000);' into any part of the Postgres code where the confirmed_flush parameter is advanced.

With this commit, we utilise standard Spock machinery to wait for the WAL flushing event, as well as to perform infinite cycles waiting for a change in subscription state.

It seems an infinite wait is not the best choice. In the future, we could introduce a 'wait_change_timeout' environment parameter that can be configured according to the characteristics of the test platform.

One more race we have detected here is that, altering the subscription, Spock just sends a signal to the apply worker. So, switching to the publisher quickly enough, we may find that the replication slot is still not ready to send data. Just wait until the plugin is ready - yeah, still in an infinite loop.

### DIFF
--- a/tests/regress/expected/interfaces.out
+++ b/tests/regress/expected/interfaces.out
@@ -50,26 +50,25 @@ SELECT * FROM spock.sub_alter_interface('test_subscription', 'super2');
 
 DO $$
 BEGIN
-    FOR i IN 1..100 LOOP
-        IF EXISTS (SELECT 1 FROM spock.sub_show_status() WHERE status != 'down') THEN
-            EXIT;
-        END IF;
-        PERFORM pg_sleep(0.1);
+    WHILE EXISTS (SELECT 1 FROM spock.sub_show_status() WHERE status = 'down')
+	LOOP
     END LOOP;
 END;$$;
-SELECT pg_sleep(0.1);
- pg_sleep 
-----------
- 
-(1 row)
-
-SELECT subscription_name, status, provider_node, replication_sets, forward_origins FROM spock.sub_show_status();
+SELECT
+  subscription_name, status, provider_node, replication_sets, forward_origins
+FROM spock.sub_show_status();
  subscription_name |   status    | provider_node |           replication_sets            | forward_origins 
 -------------------+-------------+---------------+---------------------------------------+-----------------
  test_subscription | replicating | test_provider | {default,default_insert_only,ddl_sql} | 
 (1 row)
 
 \c :provider_dsn
+DO $$
+BEGIN
+    WHILE EXISTS (SELECT 1 FROM pg_replication_slots WHERE active = 'false')
+	LOOP
+    END LOOP;
+END;$$;
 SELECT plugin, slot_type, active FROM pg_replication_slots;
     plugin    | slot_type | active 
 --------------+-----------+--------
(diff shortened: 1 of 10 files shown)


tests/regress/expected/interfaces.out
```diff
@@ -50,26 +50,25 @@ SELECT * FROM spock.sub_alter_interface('test_subscription', 'super2');
 
 DO $$
 BEGIN
-    FOR i IN 1..100 LOOP
-        IF EXISTS (SELECT 1 FROM spock.sub_show_status() WHERE status != 'down') THEN
-            EXIT;
-        END IF;
-        PERFORM pg_sleep(0.1);
+    WHILE EXISTS (SELECT 1 FROM spock.sub_show_status() WHERE status = 'down')
+	LOOP
```
mason-sharp, line 54: I think I did something like this once.. the job hung until it was eventually killed on github (an hour later?). I would rather use some maximum number of attempts and fail. Feel free to change the sleep time and/or count though.

danolivo: The maximum number of attempts is an evil approach for multimaster - there are numerous build farms, networks with varying latencies, and especially Docker, which may suspend a process for seconds.
As I wrote in the comments, it makes sense to invent a timeout to set it from the build farm side. However, for now, it would be better to simply reduce the GitHub Actions timeout to 10 minutes, I suppose.

mason-sharp: This is just for running tests.

While `timeout-minutes` sounds like a good idea, sometimes this job takes more than 10 minutes, something like 15-20 minutes would be safer. Still, if we know a particular place should wait not more than a few seconds and we generously increase that to say, 60, we will be able fail sooner than waiting the entire length of time.

danolivo: I think it is a characteristic of specific buildfarm. If standard server takes 15 minutes to sync an empty table - there are something that we don't want, Am I wrong?
So, fixed number of iterations + any pulled out of thin air time delay inside the iteration - is an evil solution. It should be something coming from environment.
However, for the time being, we only have one test platform. An approach involving infinite waiting and a timeout for the entire testing process should suffice.

mason-sharp: I would rather fail sooner.

tests/regress/expected/interfaces.out
```diff
@@ -50,26 +50,25 @@ SELECT * FROM spock.sub_alter_interface('test_subscription', 'super2');
 
 DO $$
 BEGIN
-    FOR i IN 1..100 LOOP
-        IF EXISTS (SELECT 1 FROM spock.sub_show_status() WHERE status != 'down') THEN
-            EXIT;
-        END IF;
-        PERFORM pg_sleep(0.1);
```
mason-sharp, line 57: I am not sure about the internal infrastructure and if they over-provision internally, and while it looks like there is no penalty with potential charges in terms of consuming more CPU time vs wall clock time, a brief sleep seems like a good thing to do to avoid being a noisy neighbor.

tests/regress/expected/interfaces.out
mason-sharp: I created this: https://github.com/pgEdge/spock/pull/256.

While I prefer to catch the problem when it occurs, I don't feel at all nearly so strongly about it as you do, so we can rely on the global timeout.

danolivo: Ok, so now we may wait until the timeout will be reached. Nice! It would allow us to stress-test Spock on something like Raspberry PI to reveal hidden bugs.

tests/regress/expected/interfaces.out
danolivo: Test should not be easy for a platform to pass. The sleep call reduces stress, that's bad,  and potentially increases the test time.

mason-sharp (approved):

PR #216, opened 2025-10-13
Introduce a transient Docker image (SPOC-250).

danolivo:
To avoid repeating the process of refreshing the OS image required for Spock testing, prepare it once, store it at ghcr.io, and reuse it to build a Docker image with a patched Postgres and Spock extension.

No automatic check of the cached entry is necessary; re-run the workflow manually in GitHub Actions if the image appears slightly stale.

TODO:
It makes sense to cache a Docker image with a patched Postgres branch for supported versions on top of a basic Docker image, as we don't frequently change our patches. There will be some checks necessary to ensure that the patch set hasn't been changed since the last image creation, as well as the 'install.py' script.


.github/workflows/cache-base-image.yml
```diff
@@ -0,0 +1,62 @@
...
+        # Codacy wants us to use full commit SHA. This is for v3
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3
+
+      - name: Debug output
+        run: |
+          echo "Lower-cased owner name is $OWNER_LC"
+          echo "GITHUB_WORKSPACE: ${GITHUB_WORKSPACE}"
+
+      # Login to GHCR
+      - uses: docker/login-action@v3
```
jason-lynch: I think we need to use commit SHAs instead of tags here and on line 53 to appease Codacy.

.github/workflows/cache-base-image.yml
```diff
@@ -0,0 +1,62 @@
+name: Update base OS image
```
jason-lynch, line 1: Are we going to use this image across multiple jobs? If not, you could keep skip publishing the image and use the `gha` cache to cut down on rebuilds.

tests/docker/Dockerfile-base.el9
```diff
@@ -0,0 +1,40 @@
...
+# Includes fresh updates of the base OS plus all extra packages needed to
+# Postgres, Spock and testing facilities.
+#
+# ##############################################################################
+
+FROM rockylinux:9
+
+ARG PGVER
+
+ENV PGVER=$PGVER
```
jason-lynch: I think you can remove this argument/env var from this file. It only gets set and used in the `step-1` image.

.github/workflows/cache-base-image.yml
```diff
@@ -0,0 +1,62 @@
...
+          OWNER: '${{ github.repository_owner }}'
+
+      - name: Set up Docker
+        # Codacy wants us to use full commit SHA. This is for v3
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3
+
+      - name: Debug output
+        run: |
+          echo "Lower-cased owner name is $OWNER_LC"
+          echo "GITHUB_WORKSPACE: ${GITHUB_WORKSPACE}"
```
jason-lynch: You could probably remove this now that you have it working.

.github/workflows/spockbench.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           echo PG_VER=${{ matrix.pgver }} >> tests/docker/pgedge.env
```
jason-lynch: It looks like this `echo PG_VER` line is unnecessary since you have the build arg below. Would removing it make it easier to switch this step to use `docker/build-push-action`?

.github/workflows/cache-base-image.yml
danolivo: My idea was to re-use it between all the products, no only in Spock. Saving of computational resources is my addiction ;).

.github/workflows/spockbench.yml
danolivo: Maybe. I've never been a person who likes yaml, docker or any automatisation stuff in general. Will try.

jason-lynch: I just noticed that the build arg is missing the `_`. If you do want to remove this, you could add the `_` to the build arg and refactor the Dockerfile, or refactor whichever tests read `PG_VER`.

danolivo: I see docker_compose.yml uses pgedge.env. I think, the author of the spockbench should decide on that part.

.github/workflows/cache-base-image.yml
danolivo: This comment is all Greek to me. I think I need time to understand what does it mean ;).

danolivo: Ok, now I know that this is not just a magic keyword but a link to a repository :))). Fixed.

.github/workflows/spockbench.yml
pct960: I did a `grep -nri "pg_ver" .` and nothing turned up in `tests/docker`. It does seem to be redundant since `PGVER` is used almost everywhere except for this echo statement.

danolivo:
> Some of this might be overkill for our needs, but separating these out will help us save on actions storage and minutes, with, of course, the nice side effect of fast test runs.

1. In our paradigm, we test not a commit, but the PR. Inventing image B means updating it on each commit to each PRed branch. Not sure we want it.
2. I tried to write a script to create image C, but it should be triggered on each commit to check that the patch set (or install.py) has changed. GitHub Actions, at the moment, is a little limited in guaranteeing that this image will be created before the next testing phase, so I just ended up with image A. It resolves our current problem.

.github/workflows/spockbench.yml
```diff
@@ -51,10 +51,9 @@ jobs:
       - name: Build docker container
         run: |
           cd ${GITHUB_WORKSPACE}/
-          echo PG_VER=${{ matrix.pgver }} >> tests/docker/pgedge.env
           docker build \
             --build-arg PGVER=${{ matrix.pgver }} \
-            -t spock -f tests/docker/Dockerfile.el9 .
+            -t spock -f tests/docker/Dockerfile-step-1.el9 .
```
mason-sharp, line 56: It looks like `docker compose up` towards the end of this file that runs spockbench will use `tests/docker/docker-compose.yml` which is using the old container image `Dockerfile.el9`? Should it be changed to use the step-1 image?

pct960:
@danolivo 
> In our paradigm, we test not a commit, but the PR. Inventing image B means updating it on each commit to each PRed branch. Not sure we want it.

Not really. We only rebuild the last image to compile and install Spock. All other images are cached. In fact, we could combine the pgEdge and (patched) Postgres images and call it `pgedge-postgres`. The base devel image and the other image are updated only as needed. In concrete terms:

- Image `test-base`, pre-built and cached: Contains development-related packages. You already have this with `Dockerfile-base.el9`
- Image `pgedge-postgres`, pre-built and cached: uses `test-base`. Contains installed pgEdge binaries along with compiled and installed patched Postgres. We could have three flavours of this image for different Postgres versions. This is basically what your TODO comment was referring to.
- Image `spock`, not cached: Uses `pgedge-postgres`. Built every time the GitHub Actions pipeline runs.

If we're not worried about the `test-base` image being reusable across other repositories, we could just have one cached image with devel binaries, pgedge binaries, and patched Postgres. The correct flavour of this cached image (based on the Postgres version) is pulled, Spock is compiled and installed, and the Spockbench tests are executed. This is the simplest option IMO.

>I tried to write a script to create image C, but it should be triggered on each commit to check that the patch set (or install.py) has changed. GitHub Actions, at the moment, is a little limited in guaranteeing that this image will be created before the next testing phase, so I just ended up with image A. It resolves our current problem.

We can configure github actions to selectively run a job if the `patches/` directory is changed. This job could rebuild the cached image if our patches change.

danolivo:
> We can configure github actions to selectively run a job if the `patches/` directory is changed

That is the key to why I have stopped here. We have no guarantee this job will be executed before any other testing job that may cause failure. It is just not flexible enough to design a proper YAML script. I personally don't like such false positives because they waste hours of scrutiny.
Maybe you propose to add to each script something like:
```
on:
  push:
    paths:
      - 'src/**'
```
Unfortunately, it must be checked at each script file, and each job in these scripts should verify that, at least, while we don't want to test each commit.

Anyway, I'm no welder when it comes to this stuff. So, if you have ideas, I would be glad to discover the schema with code samples - Miro diagrams are a good place for such things.

.github/workflows/spockbench.yml
danolivo: Done

mason-sharp:
One last thing- I think we should delete Dockerfile.el9 if it is no longer being used.

PR #224, opened 2025-10-17
Add checkpoint hook integration for group resource and WAL progress

rasifr:
This patch introduces a minimal checkpoint hook integration in PostgreSQL (via patch 'pg17-040-checkpoint-hook.diff') and wires it into Spock’s initialization flow.

During each checkpoint, after buffers have been flushed and before the fsync/sync phase, the new hook is invoked. Spock adds its own hook ('spock_checkpoint_hook') at this stage to:

1. Dump group data ('spock_group_resource_dump()') to disk, ensuring the latest in-memory state is persisted.

2. Emit per-group progress records to WAL using 'spock_group_foreach()' and a new callback 'spock_group_emit_progress_wal_cb()', which calls 'spock_apply_progress_add_to_wal()' for each entry.

This ensures that after a crash or restart, the replication progress can be cleanly reloaded from the last checkpoint or from the resouce file.

### DIFF
--- a/src/spock_group.c
+++ b/src/spock_group.c
@@ -65,24 +65,6 @@ static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 HTAB	   *SpockGroupHash = NULL;
 
 /*
- * Install hooks to request shared resources for apply workers
- */
-void
-spock_group_shmem_init(void)
-{
-#if 0
-#if PG_VERSION_NUM < 150000
-	spock_group_shmem_request();
-#else
-	prev_shmem_request_hook = shmem_request_hook;
-	shmem_request_hook = spock_group_shmem_request;
-#endif
-	prev_shmem_startup_hook = shmem_startup_hook;
-	shmem_startup_hook = spock_group_shmem_startup;
-#endif
-}
-
-/*
  * spock_group_shmem_request
  *
  * Request and initialize the shmem structures backing the group registry.
@@ -352,20 +334,10 @@ dump_one_group_cb(const SpockGroupEntry *e, void *arg)
  * spock_group_resource_dump
  *
  * Write a clean-shutdown snapshot to PGDATA/spock/resource.dat.
- * - Header: version, system_identifier, flags, entry_count (patched after scan)
- * - Body:   array of SpockApplyProgress records (struct layout is prefix-stable)
+ * - Header: version, system_identifier, flags, entry_count
+ * - Body:   array of SpockApplyProgress records
  * Writes to a temp file, fsyncs, then durable_rename() into place.
  * Typically invoked via on_shmem_exit() from the main Spock process.
- */
-/*
- * spock_group_resource_dump
- * -------------------------
- * Write a clean-shutdown snapshot to PGDATA/spock/resource.dat.
- * - Header: version, system_identifier, flags, entry_count (patched after scan)
- * - Body:   array of SpockApplyProgress records (struct layout is prefix-stable)
- * Writes to a temp file, fsyncs, then durable_rename_excl() into place.
- * Typically invoked via on_shmem_exit() from the main Spock process.
- *
  */
 void
 spock_group_resource_dump(void)
--- a/include/spock_group.h
+++ b/include/spock_group.h
@@ -88,7 +88,6 @@ typedef struct SpockGroupEntry
 } SpockGroupEntry;
 
 /* shmem setup */
-void		spock_group_shmem_init(void);
 extern void spock_group_shmem_request(void);
 extern void spock_group_shmem_startup(int napply_groups, bool found);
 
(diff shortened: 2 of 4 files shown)


patches/15/pg15-040-checkpoint-hook.diff
```diff
@@ -0,0 +1,52 @@
...
+@@ -7517,6 +7524,14 @@ CheckPointGuts(XLogRecPtr checkPointRedo, int flags)
+ 	CheckPointPredicate();
+ 	CheckPointBuffers(flags);
+ 
++	/*
++	 * Spock:
++	 * Extension hook: between buffer flush and fsync/sync phase.
++	 */
++	if (CheckpointHook &&
++		(flags & (CHECKPOINT_IS_SHUTDOWN | CHECKPOINT_END_OF_RECOVERY)) == 0)
```
danolivo: I think, here you just follow the common practice and pass both parameters (XLogRecPtr checkPointRedo, int flags) as is to the hook.

src/spock_group.c
```diff
@@ -500,3 +500,20 @@ spock_group_resource_load(void)
...
+	Assert((flags & (CHECKPOINT_IS_SHUTDOWN | CHECKPOINT_END_OF_RECOVERY)) == 0);
+
+	/* Dump group progress to resource.dat */
+	spock_group_resource_dump();
+
+	/* Emit progress WAL records for all groups */
+	LWLockAcquire(SpockCtx->apply_group_master_lock, LW_SHARED);
+
+	/* write all entries to WAL */
+	spock_group_foreach(spock_group_emit_progress_wal_cb, NULL);
```
danolivo: If you already successfully put all the records into persistent file on disk, what is the reason to write them to the WAL?

patches/18/pg18-040-checkpoint-hook.diff
```diff
@@ -0,0 +1,52 @@
...
+diff --git a/src/include/access/xlog.h b/src/include/access/xlog.h
+index 2c507ea618c..b76830ebcc7 100644
+--- a/src/include/access/xlog.h
++++ b/src/include/access/xlog.h
+@@ -196,6 +196,14 @@ typedef enum WALAvailability
+ struct XLogRecData;
+ struct XLogReaderState;
+ 
++/*
++ * Spock:
```
danolivo: I propose to not mention Spock in the patch. Just to be closer to the community version.

patches/18/pg18-040-checkpoint-hook.diff
```diff
@@ -0,0 +1,52 @@
...
+@@ -156,6 +156,13 @@ int			wal_segment_size = DEFAULT_XLOG_SEG_SIZE;
+  */
+ int			CheckPointSegments;
+ 
++/*
++ * Spock:
++ * Hook called during checkpoint processing, after buffers have been
++ * flushed, but before fsync/sync is done.
++ */
++CheckpointHookType CheckpointHook = NULL;
```
danolivo: It seems picky, but I propose to use naming convention more proper for the core hook name and type. See, for example, ExecutorFinish_hook_type.

src/spock_group.c
```diff
@@ -64,24 +64,6 @@ static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
 
 HTAB	   *SpockGroupHash = NULL;
 
-/*
- * Install hooks to request shared resources for apply workers
- */
-void
-spock_group_shmem_init(void)
-{
-#if 0
```
danolivo, line 73: @mason-sharp , I guess it needs your decision: this staff is not necessary because the previous RM commit removed much of the apply group code. So, if my understanding is correct and we are going to start from scratch, it makes sense. If not, it would be better to return the old code.

mason-sharp: That code has #if 0 there. I asked Asif, he said that code should be removed.

patches/15/pg15-040-checkpoint-hook.diff
rasifr: Sounds good, will update the PR.

src/spock_group.c
rasifr: I think writing both would provide more safety. Between two checkpoints, crash recovery will alway be done from WAL replay, which will start from the last checkpoint. So the reply should find all existing entries after the last checkpoint.

src/spock_group.c
rasifr: yeah so this code was marked for removal. Initially spock_group_shmem_init was added to register shmem hooks for group apply. But then changed it to use the existing init function available in spock for the same purpose. This was done to keep all hook registration central and not to spread in multiple files.

src/spock_group.c
danolivo: Not sure I understand your point. If file dumping will be crashed - you have WAL records stored before, checkpoint has not finished yet, right? If you dump it - any restore will start from this checkpoint, right? you will need this file and new updates, coming with records, updated after checkpoint.

PR #227, opened 2025-10-20
SPOC-129: Implement automatic recovery slot creation for catastrophic failure recovery.

ibrarahmad:
Add recovery slot functionality to preserve WAL segments for catastrophic node failure recovery. Each database automatically creates and maintains one shared recovery slot that enables data recovery without requiring cluster rebuild.

Each database has one inactive logical replication slot (spk_recovery_{dbname}) that preserves WAL for all peer subscriptions. The slot is created by the manager worker at database initialization and is shared across all subscriptions on that database.

Recovery slots are inactive by default and never used by normal replication. They can be monitored via spock.get_recovery_slot_status() and recreated if missing using spock.quick_health_check().

### DIFF
--- a/src/spock.c
+++ b/src/spock.c
@@ -135,6 +135,7 @@ int		restart_delay_default;
 int		restart_delay_on_exception;
 int		spock_replay_queue_size;  /* Deprecated - no longer used */
 bool	check_all_uc_indexes = false;
+bool	spock_enable_recovery_slots = true;
 
 static emit_log_hook_type prev_emit_log_hook = NULL;
 
@@ -1154,6 +1155,15 @@ _PG_init(void)
 							 0,
 							 NULL, NULL, NULL);
 
+	DefineCustomBoolVariable("spock.enable_recovery_slots",
+							 gettext_noop("Enable automatic creation of recovery slots for catastrophic failure recovery."),
+							 gettext_noop("When enabled, Spock automatically creates recovery slots to preserve WAL for catastrophic failure recovery."),
+							 &spock_enable_recovery_slots,
+							 true,
+							 PGC_SIGHUP,
+							 0,
+							 NULL, NULL, NULL);
+
 	if (IsBinaryUpgrade)
 		return;
 
@@ -1168,9 +1178,6 @@ _PG_init(void)
 
 	/* Init output plugin shmem */
 	spock_group_shmem_init();
-
-	/* Init recovery slots shmem */
-	spock_recovery_shmem_init();
 
 	/* Init executor module */
 	spock_executor_init();
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -86,7 +86,6 @@
 #include "spock_exception_handler.h"
 #include "spock_common.h"
 #include "spock_readonly.h"
-#include "spock_recovery.h"
 #include "spock.h"
 
 
--- a/src/spock_functions.c
+++ b/src/spock_functions.c
@@ -166,7 +166,6 @@ PG_FUNCTION_INFO_V1(spock_max_proto_version);
 
 /* Recovery Slot functions */
 PG_FUNCTION_INFO_V1(spock_get_recovery_slot_status_sql);
-PG_FUNCTION_INFO_V1(spock_create_recovery_slot_sql);
 
 PG_FUNCTION_INFO_V1(spock_xact_commit_timestamp_origin);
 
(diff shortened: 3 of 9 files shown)


tests/tap/schedule
```diff
@@ -1,15 +1,5 @@
...
-test: 002_create_subscriber
-
-# Replication tests
-test: 003_cascade_replication
-test: 004_non_default_repset
-
-# Advanced functionality tests
-# test: 005_daylight_savings
-# test: 006_sync_during_write
-test: 008_rmgr
```
mason-sharp, line 15: We should not delete running all of the existing tests.

src/spock_manager.c
```diff
@@ -206,6 +207,36 @@ spock_manager_main(Datum main_arg)
...
+			elog(LOG, "created recovery slot '%s' for database %s",
+				 slot_name, get_database_name(MyDatabaseId));
+			pfree(slot_name);
+		}
+	}
+	PG_CATCH();
+	{
+		elog(WARNING, "failed to create recovery slot for database %s",
+			 get_database_name(MyDatabaseId));
+		/* Continue anyway - recovery slot is optional */
```
mason-sharp: Would there be a case were it should be optional?  If optional should we make it configurable? If set to be required, then error out here? If not configured, don't bother creating it?

tests/tap/schedule
mason-sharp: Understood, these cases may fail because of slot checks, we can disable temporarily for this development branch.

src/spock.c
```diff
@@ -1168,6 +1169,9 @@ _PG_init(void)
 	/* Init output plugin shmem */
 	spock_group_shmem_init();
 
+	/* Init recovery slots shmem */
+	spock_recovery_shmem_init();
+
```
rasifr: I think we should keep all shmem hook assignments in a single place for consistency. spock_worker_shmem_init() already handles that, so instead of registering your own hooks here, I think its better to integrate with the existing ones.

 `spock_recovery_shmem_size()` can be called from `spock_worker_shmem_request()` and `spock_recovery_shmem_startup()` from `spock_worker_shmem_startup()`.

That should keeps all related shmem setup logic in sync and avoids wiring hooks across multiple init paths.

src/spock_apply.c
```diff
@@ -86,6 +86,7 @@
 #include "spock_exception_handler.h"
 #include "spock_common.h"
 #include "spock_readonly.h"
+#include "spock_recovery.h"
```
rasifr: There doesn't seem to be any changes in the file. So not sure, if this is needed.

src/spock_functions.c
```diff
@@ -3358,3 +3363,102 @@ get_apply_group_progress(PG_FUNCTION_ARGS)
...
+spock_create_recovery_slot_sql(PG_FUNCTION_ARGS)
+{
+	char	   *database_name;
+	bool		result;
+	
+	if (PG_ARGISNULL(0))
+		PG_RETURN_NULL();
+		
+	database_name = text_to_cstring(PG_GETARG_TEXT_PP(0));
+
```
rasifr: There should not be any need for getting database name from user. `MyDatabaseId` can be used here instead.

src/spock_manager.c
```diff
@@ -206,6 +207,36 @@ spock_manager_main(Datum main_arg)
...
+	 * a peer node fails without requiring cluster rebuild.
+	 *
+	 * Created here because manager runs once per database, ensuring a single
+	 * shared slot for all subscriptions.
+	 */
+	StartTransactionCommand();
+	PG_TRY();
+	{
+		if (create_recovery_slot(get_database_name(MyDatabaseId)))
+		{
```
rasifr, line 226: It seems like we’re offering multiple ways to create recovery slots.
I would suggest simplifying this by introducing a GUC to enable or disable recovery slot creation, instead of exposing a user interface.

This keeps control in the user's hands but reduces much complexity; fewer code paths to maintain, and less chance of user error or conflicting configurations.

src/spock_recovery.c
```diff
@@ -0,0 +1,489 @@
...
+	PG_TRY();
+	{
+		ReplicationSlotCreate(slot_name, true, RS_PERSISTENT, false, false, false);
+		success = true;
+
+		elog(LOG, "created recovery slot '%s' (INACTIVE - for catastrophic failure recovery only)",
+			 slot_name);
+	}
+	PG_CATCH();
+	{
```
rasifr: Let’s fail on error here. If recovery slots are enabled, failure to create one should be treated as a fatal error and result in an exit.

Using PG_TRY/CATCH just to swallow the failure complicates things unnecessarily; we can drop that and keep the logic simple.

Also, only touch shared-memory state after a successful create; that removes the success / cleanup_needed flags and the reset path entirely.

src/spock_recovery.c
```diff
@@ -0,0 +1,489 @@
...
+		}
+
+		/*
+		 * Track the restart LSN (earliest position needed).
+		 * This would be updated by WAL advancement logic.
+		 */
+		if (slot->restart_lsn == InvalidXLogRecPtr ||
+			lsn < slot->restart_lsn)
+		{
+			slot->restart_lsn = lsn;
```
rasifr, line 435: Assuming lsn is either commit lsn or end lsn here; Would it ever be less then restart_lsn?

PR #233, opened 2025-10-21
Fix two bugs in the table re-sync routine (spoc-263).

danolivo:
At first, we need to switch off WAL sending machinery when truncating and re-synchronising the table. It prevents us from losing data across the instances.

The second issue is related to the infinite waiting if no more DML is committed to the table. The SYNC apply worker report status only on commit. So, the applying worker, who has initiated this synchronisation, should provide the SYNC worker not with the last received record LSN, but with the last committed LSN.

A simple regression test set is provided.

Being in the context, revisit the documentation and visual representation of the re-sync routine in the UI script.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -3279,10 +3279,20 @@ process_syncing_tables(XLogRecPtr end_lsn)
 										 NameStr(sync->nspname),
 										 NameStr(sync->relname));
 
+				/*
+				 * TODO:
+				 * What if the SYNC worker has gone? It may be any trivial
+				 * ERROR - memory allocation, or network connection, for
+				 * example. We need to restart syncing process from the scratch.
+				 * For now, suppose that it should be always exist in testing
+				 * environment.
+				 */
+				Assert(spock_worker_running(worker));
+
 				if (spock_worker_running(worker) &&
-					end_lsn >= worker->worker.apply.replay_stop_lsn)
+					replorigin_session_origin_lsn >= worker->worker.apply.replay_stop_lsn)
 				{
-					worker->worker.apply.replay_stop_lsn = end_lsn;
+					worker->worker.apply.replay_stop_lsn = replorigin_session_origin_lsn;
 					sync->status = SYNC_STATUS_CATCHUP;
 
 					StartTransactionCommand();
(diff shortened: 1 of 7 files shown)


src/spock_apply.c
```diff
@@ -3279,10 +3279,19 @@ process_syncing_tables(XLogRecPtr end_lsn)
...
+				 * For now, suppose that it should be always exist in testing
+				 * environment.
+				 */
+				Assert(spock_worker_running(worker));
+
 				if (spock_worker_running(worker) &&
-					end_lsn >= worker->worker.apply.replay_stop_lsn)
+					replorigin_session_origin_lsn >= worker->worker.apply.replay_stop_lsn)
 				{
-					worker->worker.apply.replay_stop_lsn = end_lsn;
```
rasifr, line 3285: It appears that `replay_stop_lsn` is only ever initialized with `replorigin_session_origin_lsn`. This suggests either the current condition is no longer valid, or `replay_stop_lsn` needs to be updated appropriately at some point.

It might be more appropriate to either remove `replay_stop_lsn` entirely, or ensure that the caller of `process_syncing_tables` provides the correct LSN for comparison.

danolivo: Not sure I understand. Apply worker may apply more transactions before the SYNC worker switches to the SYNCWAIT state. So, replorigin_session_origin_lsn of apply_worker may be far away from the SYNCer. Do I understand this logic correctly?

rasifr: I see. Yeah that makes sense. Since replorigin_session_origin_lsn can advance during state changes; Is it possible that the next condition also needs to reflect that? i.e.

```
			if (sync->status == SYNC_STATUS_SYNCDONE &&
				end_lsn >= sync->statuslsn)
```

end_lsn here could be a stale lsn, So this condition might fail and won't update sync status. Would it be safer to compare against live replorigin_session_origin_lsn instead?

danolivo: In principle it makes sense. But mind the following:
The SYNC worker replay_stop_lsn points to the end of target COMMIT command. That's the reason why we identify that the 'copy' stage has ended (we acquired all the transactions) and we can switch to the catchup stage.
But APPLY workers works differently and write to the sync_status LSN of different nature - usually the LSN of last written to the WAL. So, they accumulate all the WAL changes, even non-transactional ones. Changing this place we need to change at the same time semantics of the `sync->status_lsn` field and I'm not sure if it is for good.

PR #238, opened 2025-10-23
SPOC-306: Fix zodan.py and add tap test

mason-sharp:
Also remove `007_zodan_node_addition.pl`, since `009_zodan_add_remove_nodes.pl` has similar coverage.


tests/tap/schedule
```diff
@@ -14,3 +14,4 @@ test: 004_non_default_repset
 # test: 006_sync_during_write
 test: 008_rmgr
 test: 009_zodan_add_remove_nodes
+test: 010_zodan_add_remove_python
```
ibrarahmad, line 17: You forgot to add 010_zodan_add_remove_python.pl.

ibrarahmad (changes requested):

tests/tap/schedule
mason-sharp: Thanks! Added.

PR #239, opened 2025-10-23
SPOC-134: Add rescue coordinator for disaster recovery.

ibrarahmad:
Implement spock.find_rescue_source() function to find the best surviving node for recovery operations when an origin node fails. Returns LSN, timestamp, and confidence level for automated recovery decisions.

### DIFF
--- a/src/spock_recovery.c
+++ b/src/spock_recovery.c
@@ -34,26 +34,29 @@
  */
 #include "postgres.h"
 
-#include "miscadmin.h"
+#include "access/heapam.h"
 #include "access/xlog.h"
 #include "catalog/namespace.h"
+#include "catalog/pg_subscription.h"
 #include "commands/dbcommands.h"
+#include "executor/spi.h"
+#include "funcapi.h"
+#include "libpq-fe.h"
+#include "miscadmin.h"
 #include "replication/slot.h"
 #include "storage/ipc.h"
 #include "storage/lwlock.h"
 #include "storage/shmem.h"
 #include "utils/builtins.h"
+#include "utils/elog.h"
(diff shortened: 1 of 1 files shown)


src/spock_recovery.c
```diff
@@ -46,9 +46,13 @@
 #include "utils/memutils.h"
 #include "utils/pg_lsn.h"
 #include "utils/timestamp.h"
+#include "access/heapam.h"
+#include "catalog/pg_subscription.h"
```
mason-sharp: Might as well try to keep these alphabetical.

src/spock_recovery.c
```diff
@@ -442,3 +446,222 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+ *
+ * Returns a record with:
+ * - origin_node_id: The failed node ID
+ * - source_node_id: The surviving node ID with the most recent data
+ * - last_lsn: The LSN of the last known transaction from the origin
+ * - last_commit_timestamp: The commit timestamp of the last known transaction
+ * - confidence_level: How confident we are in this choice (HIGH/MEDIUM/LOW)
+ */
+Datum
+spock_find_rescue_source_sql(PG_FUNCTION_ARGS)
```
mason-sharp, line 578: Not sure where I see where it is querying the other nodes. That is, if n4 goes down, I thought n1 would query n2 and n3 to see which one has received more transactions from n4.

ibrarahmad: I missed the last commit, done.

src/spock_recovery.c
ibrarahmad: fixed

PR #241, opened 2025-10-23
SPOC-212: document exception_replay_queue_size

mason-sharp:
Recently deprecated, describe new handling

### DIFF
--- a/docs/install_spock.md
+++ b/docs/install_spock.md
@@ -187,10 +187,6 @@ Use this GUC to specify which operations/transactions are written to the excepti
 * `discard` - Add a row to the `spock.exception_log` table for any discarded operation; successful transactions are not logged.
 * `none` - Instructs the server to not log any operation or transactions to the exception log table.
 
-#### `spock.exception_replay_queue_size`
-
-When Spock encounters a replication exception, it attempts to resolve the exception by entering exception-handling mode, based on the value of `spock.exception_behaviour`.  Spock then writes any transaction up to a default size of `4MB` to memory, and the apply worker replays the transaction from memory.  This provides a massive speed and performance increase in the handling of the vast majority of exceptions.  The memory size is configurable with `spock.exception_replay_queue_size`.
-
 #### `spock.extra_connection_options`
 
 You can use the `spock.extra_connection_options` parameter in the `postgresql.conf` file to assign connection options that apply to all connections made by Spock. This can be a useful place to set up custom keepalive options, etc.
--- a/docs/spock_release_notes.md
+++ b/docs/spock_release_notes.md
@@ -1,5 +1,13 @@
 # Spock Release Notes
 
+## Spock 5.1 on xxx
+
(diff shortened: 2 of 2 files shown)


docs/install_spock.md
```diff
@@ -187,10 +187,6 @@ Use this GUC to specify which operations/transactions are written to the excepti
 * `discard` - Add a row to the `spock.exception_log` table for any discarded operation; successful transactions are not logged.
 * `none` - Instructs the server to not log any operation or transactions to the exception log table.
 
-#### `spock.exception_replay_queue_size`
```
ibrarahmad, line 190: @susan-pgedge, do you think we should keep that here and mention that it's deprecated?

susan-pgedge: Probably for at least one version...

docs/spock_release_notes.md
```diff
@@ -1,5 +1,13 @@
 # Spock Release Notes
 
+## Spock 5.1 on xxx
+
+* Deprecate the spock.exception_replay_queue_size GUC. Previously spock would store transaction changes up to the size defined by the spock.exception_replay_queue_size GUC. If an error occurred, the transaction was replayed, and the size was less than this amount, the cache would was used. If it was greater than this limit, it would be resent from the origin. Now no restriction exists, it will use memory until exhasted (helping performance for huge transactions). If an allocaiton fails, it will follow the spock.exception_behavior GUC:
+    - 'discard': Skip failed transaction and continue
+    - 'transdiscard': Rollback transaction and continue
+    - 'sub_disable': Disable subscription and exit cleanly
+
+
```
susan-pgedge, line 10: Some grammar/spelling updates:

This release deprecates the spock.exception_replay_queue_size GUC. Previously Spock restored transaction changes up to the size defined by the spock.exception_replay_queue_size GUC. If an error occurred, the transaction was replayed, and if the size was less than the exception queue, the cache was used. If the size was greater than the limit, it was resent from the origin. 

Now no restriction exists.  Spock will use memory until memory is exhausted (improving performance for huge transactions). If an allocation fails, Spock performs as specified by the spock.exception_behavior GUC:
    - 'discard': Skip failed transaction and continue
    - 'transdiscard': Rollback transaction and continue
    - 'sub_disable': Disable subscription and exit cleanly

docs/install_spock.md
```diff
@@ -187,10 +187,6 @@ Use this GUC to specify which operations/transactions are written to the excepti
...
-
-When Spock encounters a replication exception, it attempts to resolve the exception by entering exception-handling mode, based on the value of `spock.exception_behaviour`.  Spock then writes any transaction up to a default size of `4MB` to memory, and the apply worker replays the transaction from memory.  This provides a massive speed and performance increase in the handling of the vast majority of exceptions.  The memory size is configurable with `spock.exception_replay_queue_size`.
-
```
susan-pgedge, line 193: #### `spock.exception_replay_queue_size` - **DEPRECATED**

This parameter will not be supported after version 5.X.

When Spock encounters a replication exception, it attempts to resolve the exception by entering exception-handling mode, based on the value of `spock.exception_behaviour`.  Spock then writes any transaction up to a default size of `4MB` to memory, and the apply worker replays the transaction from memory.  This provides a massive speed and performance increase in the handling of the vast majority of exceptions.  The memory size is configurable with `spock.exception_replay_queue_size`.

Now, Spock performs as specified by the [spock.exception_behavior](ADD_LINK_TO_GUC_PAGE) parameter.

PR #243, opened 2025-10-24
Resource Manager Review (SPOC-309)

danolivo:
By this series of commits we revise out progress tracking system a little:

1. Rename last_received_lsn to commit_lsn and change behaviour correspondingly.
2. Introduce received_lsn to track the most recent lsn arrived from the publisher
3. Update remote_insert_lsn on each WAL record inbound except 'keepalive' messages because it means breaking the core part of the LR protocol.

With all these changes:
1. commit_lsn allows to identify precisely specific commit in the WAL of the publisher which was applied.
2. (remote_insert_lsn - commit_lsn) - how far away applied commit from the end of the publisher's WAL.
3. (remote_insert_lsn  - received_lsn) - replication lag.
4. (received_lsn - commit_lsn) may detect anomalies like massive transaction or DML misbalance between nodes.

### DIFF
--- a/include/spock_group.h
+++ b/include/spock_group.h
@@ -90,7 +90,8 @@ typedef enum
  * the COMMIT apply if only all other commits with smaller timestamps have
  * already been committed by other workers. So, this value tells us about
  * the real progress.
- * prev_remote_ts - obsolete value. TODO: Should be removed in further commits.
+ * prev_remote_ts - XXX: It seems do nothing at the moment and should
+ * be considered to be removed.
  * remote_commit_lsn - LSN of the COMMIT corresponding to the remote_commit_ts.
  * remote_insert_lsn - an LSN of the most advanced WAL record written to
  * the WAL on the remote side. Replication protocol attempts to update it as
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -241,9 +241,6 @@ static void spock_apply_worker_detach(void);
 
 static bool should_log_exception(bool failed);
 
-static void check_and_update_progress(XLogRecPtr last_received_lsn,
-								TimestampTz timestamp);
-
 static ApplyReplayEntry *apply_replay_entry_create(int r, char *buf);
 static void apply_replay_entry_free(ApplyReplayEntry *entry);
 static void apply_replay_queue_reset(void);
@@ -2826,9 +2823,6 @@ stream_replay:
 					 */
 					UpdateWorkerStats(last_received, last_inserted);
 
-					if (reply_requested)
-						check_and_update_progress(endpos, GetCurrentTimestamp());
-
 					/* We do not add 'k' messages to the replay queue */
 					apply_replay_entry_free(entry);
 				}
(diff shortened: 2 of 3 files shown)


utils/spockctrl/spockctrl.json
```diff
@@ -1,45 +0,0 @@
-{
```
mason-sharp, line 1: Please don't delete this.

include/spock_group.h
```diff
@@ -62,6 +62,47 @@ typedef struct SpockGroupKey
...
+ * the WAL on the remote side. Replication protocol attempts to update it as
+ * frequently as possible, but it still be a little stale.
+ * received_lsn - an LSN of the most advanced WAL record that was received by
+ * the group.
+ * last_updated_ts - timestamp when remote COMMIT command (identified by the
+ * remote_commit_ts and remote_commit_lsn) was applied locally.
+ * Spock employs this value to calculate replication_lag.
+ * updated_by_decode - obsolete value. It was needed to decide on the LR lag
+ * that seems not needed if we have NULL value for a timestamp column.
+ */
```
mason-sharp, line 106: 👍

docs/monitoring/lag_tracking.md
```diff
@@ -9,7 +9,7 @@ Tracking is implemented via a SQL view named `spock.lag_tracker`, which exposes
 origin_name           | node1
 receiver_name         | node2
 commit_timestamp      | 2025-06-30 09:27:57.317779+00
-last_received_lsn     | 0/15A2780
+commit_lsn            | 0/15A2780
```
mason-sharp, line 12: Glad you included doc updates. Please also add to docs/spock_release_notes.md. Since you want to change column names, please point it out in the description of the PR, too. (Any API/interface changes should be highlighted.)

danolivo: There is no clear protocol how to add into spock_release_notes.md. I see a section for the 5.0.4 only.

src/spock_apply.c
```diff
@@ -2772,6 +2812,19 @@ apply_work(PGconn *streamConn)
...
+					 * It allows Spock to maintain LSN lag statistic.
+					 *
+					 * NOTE: we can't change the keepalive message format, so
+					 * just apply the same last_inserted. It may cause negative
+					 * delta, but it seems not important.
+					 */
+					UpdateWorkerStats(last_received, last_inserted);
+
 					if (reply_requested)
 						check_and_update_progress(endpos, GetCurrentTimestamp());
```
rasifr: This looks suspicious to me. These functions UpdateWorkerStats and check_and_update_progress write remote_insert_lsn. Two writers on the same field can undo each other, interfering the progress. I suggest to have a single writer, maybe combine them both.

include/spock_group.h
```diff
@@ -62,6 +62,47 @@ typedef struct SpockGroupKey
...
+
+/*
+ * Logical Replication Progress has made by a group of apply workers.
+ *
+ * remote_commit_ts - the most advanced timestamp of COMMIT commands, already
+ * applied by the replication group. In fact, an apply worker may finish
+ * the COMMIT apply if only all other commits with smaller timestamps have
+ * already been committed by other workers. So, this value tells us about
+ * the real progress.
+ * prev_remote_ts - obsolete value. TODO: Should be removed in further commits.
```
rasifr: This field is for internal use only and shouldn't be exposed to user or used by anyone. So lets document it as such, instead of marking it obsolete.

src/spock_apply.c
danolivo: Sure! The check_and_update_progress function seems strange for me. I just was'n so brave to remove the code I don't fully understand.

include/spock_group.h
danolivo: Sorry, I don't understand. Do you recommend to remove it right now?

rasifr: yeah either remove it from comments, or say "internal use..." (instead of obsolete)

danolivo: No, I mean, remove this field at all. I have written 'obsolete' because I don't see any real need for it. So, it seems correct. It means we need to remove it shortly when we will be sure it actually doesn't provide any benefit.

src/spock_apply.c
danolivo: Ok, I removed check_and_update_progress

include/spock_group.h
danolivo: Replaced the comment. Formally this field is still used in the code.

danolivo:
@mason-sharp , I finalised changes. You may continue review.

PR #249, opened 2025-10-28
SPOC-134: Add rescue coordinator for disaster recovery.

ibrarahmad:
Implement spock.find_rescue_source() function to find the best surviving node for recovery operations when an origin node fails. Returns LSN, timestamp, and confidence level for automated recovery decisions.

### DIFF
--- a/src/spock_recovery.c
+++ b/src/spock_recovery.c
@@ -500,32 +500,25 @@ query_node_recovery_progress(const char *node_dsn, const char *origin_node_name,
 
 			if (lsn_str && strlen(lsn_str) > 0)
 			{
-				uint32		xlogid;
-				uint32		xrecoff;
+				*remote_lsn = str_to_lsn(lsn_str);
 
-				/* Parse LSN from string format (X/X) */
-				if (sscanf(lsn_str, "%X/%X", &xlogid, &xrecoff) == 2)
+				/* Parse timestamp if available */
+				if (ts_str && strlen(ts_str) > 0 && !PQgetisnull(res, 0, 1))
 				{
-					*remote_lsn = ((uint64) xlogid << 32) | xrecoff;
-
-					/* Parse timestamp if available */
-					if (ts_str && strlen(ts_str) > 0 && !PQgetisnull(res, 0, 1))
-					{
-						/* Convert ISO timestamp string to TimestampTz */
-						Datum		ts_datum;
-						ts_datum = DirectFunctionCall3(timestamptz_in,
-													  CStringGetDatum(ts_str),
-													  ObjectIdGetDatum(InvalidOid),
-													  Int32GetDatum(-1));
-						*remote_ts = DatumGetTimestampTz(ts_datum);
-					}
-					else
-					{
-						*remote_ts = 0;
-					}
-
-					success = true;
+					/* Convert ISO timestamp string to TimestampTz */
+					Datum		ts_datum;
+					ts_datum = DirectFunctionCall3(timestamptz_in,
+												  CStringGetDatum(ts_str),
+												  ObjectIdGetDatum(InvalidOid),
+												  Int32GetDatum(-1));
+					*remote_ts = DatumGetTimestampTz(ts_datum);
 				}
+				else
+				{
+					*remote_ts = 0;
+				}
+
+				success = true;
 			}
 		}
 		else if (PQresultStatus(res) != PGRES_TUPLES_OK)
@@ -756,14 +749,11 @@ spock_find_rescue_source_sql(PG_FUNCTION_ARGS)
 	if (OidIsValid(best_source_node_id))
 	{
 		SpockNode *source_node = get_node(best_source_node_id);
-		char	   *source_node_name = source_node ? source_node->name : "unknown";
-		char	   *lsn_str = psprintf("%X/%X", LSN_FORMAT_ARGS(best_lsn));
+		char	   *source_node_name = source_node->name;
 		const char *ts_str = timestamptz_to_str(best_commit_ts);
 
-		elog(LOG, "Rescue source for failed node %s is node %s at commit timestamp %s / LSN %s (confidence: %s)",
-			 failed_node_name, source_node_name, ts_str, lsn_str, confidence_level);
-
-		pfree(lsn_str);
+		elog(LOG, "Rescue source for failed node %s is node %s at commit timestamp %s / LSN %X/%X (confidence: %s)",
+			 failed_node_name, source_node_name, ts_str, LSN_FORMAT_ARGS(best_lsn), confidence_level);
 	}
 	else
 	{


src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+		initStringInfo(&query);
+		appendStringInfo(&query,
+			"SELECT pos.remote_lsn, "
+			"       CASE WHEN pos.remote_lsn = '0/0'::pg_lsn THEN NULL "
+			"            ELSE pg_catalog.pg_xact_commit_timestamp(pos.local_id) "
+			"       END as remote_timestamp "
+			"FROM spock.node n "
+			"JOIN pg_replication_origin_status pos "
+			"  ON pos.external_id = spock.spock_origin_name(n.node_id) "
+			"WHERE n.node_name = $1");
```
rasifr, line 489: I think spock.progress should be quried here instead;

src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+			char	   *ts_str = PQgetvalue(res, 0, 1);
+
+			if (lsn_str && strlen(lsn_str) > 0)
+			{
+				uint32		xlogid;
+				uint32		xrecoff;
+
+				/* Parse LSN from string format (X/X) */
+				if (sscanf(lsn_str, "%X/%X", &xlogid, &xrecoff) == 2)
+				{
```
rasifr: maybe use pg_lsn_in?

src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+	/* Get node ID for the failed node */
+	{
+		SpockNode *node = get_node_by_name(failed_node_name, true);
+		if (node == NULL)
+		{
+			/* Return empty result for non-existent nodes */
+			pfree(failed_node_name);
+			PG_RETURN_NULL();
+		}
+		failed_node_id = node->id;
```
rasifr, line 597: If we don't find the failed node in our system, should we simply bail out completely? I think we should and raise error, letting caller know.

src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+				if (isnull)
+					continue;
+
+				node_name = SPI_getvalue(tuple, tupdesc, 2);
+				if (!node_name)
+					continue;
+
+				node_dsn = SPI_getvalue(tuple, tupdesc, 3);
+				if (!node_dsn)
+					continue;
```
rasifr, line 648: any of these value having NULL means we have some corruption of data or system is not configured and should be considered as failure

src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+	}
+
+	if (best_lsn != InvalidXLogRecPtr)
+	{
+		values[2] = LSNGetDatum(best_lsn);
+		nulls[2] = false;
+	}
+
+	if (best_commit_ts != 0)
+	{
```
rasifr, line 738: if source_count = 0, that should be considered error and we don't need to proceed further. and when we do, best_* are assumed to be properly initilied. they also cann't have invalid or 0 because that again would mean there is something wrong in setup and a valid condition for error out.

src/spock_recovery.c
```diff
@@ -442,3 +449,329 @@ update_recovery_slot_progress(const char *slot_name, XLogRecPtr lsn,
...
+
+	tuple = heap_form_tuple(tupdesc, values, nulls);
+
+	/* Log the rescue decision */
+	if (OidIsValid(best_source_node_id))
+	{
+		SpockNode *source_node = get_node(best_source_node_id);
+		char	   *source_node_name = source_node ? source_node->name : "unknown";
+		char	   *lsn_str = psprintf("%X/%X", LSN_FORMAT_ARGS(best_lsn));
+		const char *ts_str = timestamptz_to_str(best_commit_ts);
```
rasifr: if we have a valid node, it must have name; similarly LSN_FORMAT_ARGS(best_lsn) can be used directly and will not require mem free

src/spock_recovery.c
ibrarahmad: Do we have the correct, updated information in the progress table?

src/spock_recovery.c
ibrarahmad: PostgreSQL Best Practices: For table-valued functions, returning an empty result set (via PG_RETURN_NULL()) is the standard way to handle "no data found" scenarios, rather than raising an error.

Caller Flexibility: The current design allows callers to handle non-existent nodes gracefully by checking for empty results, rather than forcing them to use exception handling.

src/spock_recovery.c
ibrarahmad: 1. Skips nodes with NULL values (continue)
2. Continues processing other nodes
3. Returns a result with NULL values for fields that couldn't be determined
4. Allows the caller to handle the "no valid sources" scenario

src/spock_recovery.c
ibrarahmad: **Why the Current Behavior is Appropriate**

**Function Contract:** The function returns a table with potentially NULL values when no valid sources are found. This is evident from:

- The test at line 140-142 explicitly expects NULL values for non-existent nodes
- The function documentation doesn't specify that it should error out

**Recovery Function Design:** This is a recovery function that should be resilient and provide information even when the situation is not ideal. The caller can decide how to handle the case when no sources are found.

**PostgreSQL Pattern:** Table-valued functions commonly return empty results or NULL values rather than raising errors when no data is available.

**Graceful Degradation:** The function provides:

- origin_node_id: Always set (the failed node)
- source_node_id: NULL if no valid sources found
- last_lsn: NULL if no valid sources found
- last_commit_timestamp: NULL if no valid sources found
- confidence_level: Always "LOW" (default)

src/spock_recovery.c
ibrarahmad: Fixed.

src/spock_recovery.c
ibrarahmad: Good point, Fixed.

src/spock_recovery.c
rasifr: Yes

src/spock_recovery.c
rasifr: The point is that, function knows the failed node. Any-other node cannot have information missing. If it does thats cause for caution and should fail. Now failure can be presented in form of raising error or returning NULL, either way is fine.

src/spock_recovery.c
rasifr: Again, Its not how we declare failure, its we should declare failure when something wrong is detected.

src/spock_recovery.c
ibrarahmad: I doubt that is why I want to query each node to see the exact value.

src/spock_recovery.c
ibrarahmad: It's the user function; if it does not find the node and returns NULL, it can figure out the issue. It's normal behavior as with other postgresql user-defined functions.

src/spock_recovery.c
ibrarahmad: It is true for other function but not for the user interactive function.

PR #250, opened 2025-10-30
SPOC-135 : Add SQL function to clone inactive recovery slot for DR workflows.

ibrarahmad:
This patch implements spock.clone_recovery_slot, which allows cloning an inactive recovery slot into a temporary, writable slot for rescue and disaster recovery operations. The cloned slot preserves the original restart_lsn and can be used for targeted WAL replay. Logging and safety requirements are maintained in accordance with standards.

### DIFF
--- a/include/spock_recovery.h
+++ b/include/spock_recovery.h
@@ -91,9 +91,4 @@ extern Datum spock_find_rescue_source_sql(PG_FUNCTION_ARGS);
  */
 extern Datum spock_clone_recovery_slot_sql(PG_FUNCTION_ARGS);
 
-/*
- * Recovery slot cloning functions
- */
-extern Datum spock_clone_recovery_slot_sql(PG_FUNCTION_ARGS);
-
 #endif							/* SPOCK_RECOVERY_H */
--- a/src/spock_recovery.c
+++ b/src/spock_recovery.c
@@ -774,7 +774,10 @@ spock_find_rescue_source_sql(PG_FUNCTION_ARGS)
  * for disaster recovery operations. The cloned slot can be used for selective
  * replay of unacknowledged transactions.
  *
- * Usage: SELECT * FROM spock.clone_recovery_slot('origin_node', 'target_node');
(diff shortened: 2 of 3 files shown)


include/spock_recovery.h
```diff
@@ -86,4 +86,14 @@ extern void update_recovery_slot_progress(const char *slot_name,
...
 
+/*
+ * Recovery slot cloning functions
+ */
+extern Datum spock_clone_recovery_slot_sql(PG_FUNCTION_ARGS);
+
+/*
+ * Recovery slot cloning functions
+ */
+extern Datum spock_clone_recovery_slot_sql(PG_FUNCTION_ARGS);
```
rasifr: duplicates?

src/spock_recovery.c
```diff
@@ -765,3 +766,187 @@ spock_find_rescue_source_sql(PG_FUNCTION_ARGS)
...
+ * - original_slot_name: Name of the original recovery slot
+ * - restart_lsn: LSN position to start replay from
+ * - success: Whether the cloning was successful
+ * - message: Status message about the operation
+ */
+Datum
+spock_clone_recovery_slot_sql(PG_FUNCTION_ARGS)
+{
+	text	   *origin_node_text = PG_GETARG_TEXT_PP(0);
+	text	   *target_node_text = PG_GETARG_TEXT_PP(1);
```
rasifr: This looks like Its trying to clone a slot from a origin node to a local node. Why do we need that?

Correct me If I'm wrong, IIRC, the idea was to clone the existing recovery slot, on the same node thats selected as source. That was to prevent the original recovery slot to change, until actual recovery is done for safety or restart purposes etc.

include/spock_recovery.h
ibrarahmad: Fixed

src/spock_recovery.c
ibrarahmad: Fixed

PR #257, opened 2025-11-03
Add spock_create_subscriber — bootstrap a subscriber from a physical basebackup

rasifr:
Adds a standalone frontend utility that:
    
    - Takes a 'pg_basebackup' from the provider and replays to a named restore point.
    - Creates per-database logical slots on the provider (optionally '--drop-slot-if-exists').
    - Starts the new node to catch up with the provider to the restore point.
    - Installs the Spock extension, creates/advances the replication origin to
      the restore-point LSN, and creates the subscription to subscribe to  the
      provider.


mason-sharp:
Closing until we have an improved solution.

PR #266, opened 2025-11-12
zodan.sql - add subscription health check (Spoc 294).

danolivo:


src/spock.c
```diff
@@ -1108,11 +1108,11 @@ _PG_init(void)
 							"This setting is deprecated and has no effect. "
 							"The replay queue now dynamically allocates memory as needed.",
 							&spock_replay_queue_size,
-							4194304,
+							4,
 							0,
-							INT_MAX,
+							MAX_KILOBYTES / 1024,
```
ibrarahmad, line 1111: What is reason to change that?

tests/tap/t/012_zodan_basics.pl
```diff
@@ -77,6 +88,40 @@
 ok($result eq '1', "Check state of the test table on N3 after the attachment");
 print STDERR "Z0DAN should add N3 to the cluster\n";
 
+# ##############################################################################
```
ibrarahmad, line 89: I don't understand the reason?

danolivo: Reason for what exactly?

danolivo: > I don't understand the reason?

Do you mean, reason for a new test or some check inside?

src/spock.c
danolivo: It is just because of boring complain in the log. We assign this GUC in terms of megabytes, so we need to declare units and proper limits too.

ibrarahmad:
Please resove conflicts

PR #272, opened 2025-11-26
Replace Postgres-based delta apply approach with extension-only one

danolivo:
No TAP tests yet. Just for the demo and check actions.

### DIFF
--- a/include/spock.h
+++ b/include/spock.h
@@ -28,7 +28,7 @@
 #define SPOCK_VERSION_NUM 60000
 
 #define EXTENSION_NAME "spock"
-#define spock_SECLABEL_PROVIDER "spock"
+#define SPOCK_SECLABEL_PROVIDER "spock"
 
 #define REPLICATION_ORIGIN_ALL	"all"
 
--- a/include/spock_relcache.h
+++ b/include/spock_relcache.h
@@ -48,8 +48,8 @@ typedef struct SpockRelation
 	/* Additional cache, only valid as long as relation mapping is. */
 	bool		hasTriggers;
 
-	Oid		   *delta_functions;
-	bool		has_delta_apply;
+	Oid		   *delta_apply_functions;
+	bool		has_delta_columns;
 } SpockRelation;
 
 extern void spock_relation_cache_update(uint32 remoteid,
--- a/src/spock.c
+++ b/src/spock.c
@@ -1200,5 +1200,5 @@ _PG_init(void)
 	emit_log_hook = log_message_filter;
 
 	/* Security label provider hook */
-	register_label_provider(spock_SECLABEL_PROVIDER, spock_object_relabel);
+	register_label_provider(SPOCK_SECLABEL_PROVIDER, spock_object_relabel);
 }
--- a/src/spock_apply_heap.c
+++ b/src/spock_apply_heap.c
@@ -602,7 +602,7 @@ build_delta_tuple(SpockRelation *rel, SpockTupleData *oldtup,
 		int			remoteattnum = rel->attmap[attidx];
 
 		Assert(remoteattnum < tupdesc->natts);
-		if (rel->delta_functions[remoteattnum] == InvalidOid)
+		if (rel->delta_apply_functions[remoteattnum] == InvalidOid)
 		{
 			deltatup->values[remoteattnum] = 0xdeadbeef;
 			deltatup->nulls[remoteattnum] = true;
@@ -630,7 +630,7 @@ build_delta_tuple(SpockRelation *rel, SpockTupleData *oldtup,
 								 &loc_isnull);
 		Assert(!loc_isnull);
 
-		result = OidFunctionCall3Coll(rel->delta_functions[remoteattnum],
+		result = OidFunctionCall3Coll(rel->delta_apply_functions[remoteattnum],
 									  InvalidOid, oldtup->values[remoteattnum],
 									  newtup->values[remoteattnum], loc_value);
 		deltatup->values[remoteattnum] = result;
@@ -755,7 +755,7 @@ spock_handle_conflict_and_apply(SpockRelation *rel, EState *estate,
 						  xmin, local_origin_found, local_origin,
 						  local_ts, idxused);
 
-	if (rel->has_delta_apply)
+	if (rel->has_delta_columns)
 	{
 		SpockTupleData deltatup;
 		HeapTuple	currenttuple;
--- a/src/spock_executor.c
+++ b/src/spock_executor.c
@@ -215,7 +215,7 @@ DeleteSecurityLabels(const char *provider)
 		Assert(!isnull);
 		provider = TextDatumGetCString(datum);
 
-		if (strcmp(provider, spock_SECLABEL_PROVIDER) != 0)
+		if (strcmp(provider, SPOCK_SECLABEL_PROVIDER) != 0)
 			continue;
 
 		CatalogTupleDelete(pg_seclabel, &htup->t_self);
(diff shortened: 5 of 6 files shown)


include/spock_relcache.h
```diff
@@ -44,11 +44,12 @@ typedef struct SpockRelation
 	Oid			idxoid;
 	Relation	rel;
 	int		   *attmap;
-	bool		has_delta_columns;
-	Oid		   *delta_apply_functions;
 
 	/* Additional cache, only valid as long as relation mapping is. */
 	bool		hasTriggers;
+
+	Oid		   *delta_functions;
```
mason-sharp: I feel we should avoid changing variable names unless the original name is very confusing. It makes diffing different versions more difficult to sort through

src/spock_apply.c
```diff
@@ -3190,9 +3190,12 @@ spock_execute_sql_command(char *cmdstr, char *role, bool isTopLevel)
 		/*
 		 * check if it's a DDL statement. we only do this for
 		 * in_spock_replicate_ddl_command
+		 * SECURITY LABEL command is not a DDL, just an utility one. Hence, let
+		 * spock execute this command.
 		 */
 		if (in_spock_replicate_ddl_command &&
-			GetCommandLogLevel(command->stmt) != LOGSTMT_DDL)
+			GetCommandLogLevel(command->stmt) != LOGSTMT_DDL &&
```
mason-sharp, line 3432: This is probably a question more for @rasifr (old code). Why would we throw an error based on the logging level? And if so, why use `!=` instead of `>=` ?

sql/spock--5.0.4--6.0.0-devel.sql
```diff
@@ -57,3 +57,115 @@ CREATE VIEW spock.lag_tracker AS
...
+
+  IF EXISTS (SELECT 1 FROM pg_catalog.pg_seclabel
+			 WHERE objoid = rel AND classoid = 'pg_class'::regclass AND
+			       provider = 'spock') THEN
+    /*
+     * Call it each time to trigger relcache invalidation callback that causes
+     * refresh of the SpockRelation entry and guarantees actual state of the
+     * delta_apply columns.
+     */
+    EXECUTE format('ALTER TABLE %I REPLICA IDENTITY FULL', rel);
```
mason-sharp: Could we just use `ALTER TABLE <table>  ALTER COLUMN <column> SET (LOG_OLD_VALUE=true)` instead of using `REPLICA IDENTITY FULL`?

include/spock.h
```diff
@@ -28,6 +28,7 @@
 #define SPOCK_VERSION_NUM 60000
 
 #define EXTENSION_NAME "spock"
+#define spock_SECLABEL_PROVIDER "spock"
```
mason-sharp: Should we use capitalized SPOCK?

sql/spock--6.0.0-devel.sql
```diff
@@ -617,3 +631,115 @@ BEGIN
...
+
+  IF EXISTS (SELECT 1 FROM pg_catalog.pg_seclabel
+			 WHERE objoid = rel AND classoid = 'pg_class'::regclass AND
+			       provider = 'spock') THEN
+    /*
+     * Call it each time to trigger relcache invalidation callback that causes
+     * refresh of the SpockRelation entry and guarantees actual state of the
+     * delta_apply columns.
+     */
+    EXECUTE format('ALTER TABLE %I REPLICA IDENTITY FULL', rel);
```
mason-sharp, line 749: If I comment out this stuff as well as the `RAISE` above, it seems to work ok.

That is, if I update a delta apply column, since the value changed, it will need to send over the value.

If I don't update the column, it does not change, it does not send the changed value, which is fine.

So, do we actually need `REPLICA IDENTITY FULL` if we only send over data when the column value is modified?

include/spock_relcache.h
danolivo: Not sure I understand what it means - it is a new variable, isn't it?

src/spock_apply.c
danolivo: I usually don't touch the code unless I 100% know how it works and why exactly this way. This question is out of scope for the current PR. Am I wrong?

include/spock.h
danolivo: We may capitalise the define's name - I just followed the naming example (borrowed from pgactive) to make cross-code browsing easier. We definitely shouldn't capitalise the define's value.

sql/spock--6.0.0-devel.sql
danolivo: The scope of this task was to reimplement the feature in an extensible manner, not to revise its code. The initial developer might have reasons, and I just followed the common idea. At least it is a much safer way to manage development risks. We can optimise the code later if needed. Of course, in this case, we would need to invent an alternative way to send invalidation messages.

include/spock_relcache.h
mason-sharp: It looked like `delta_apply_functions` -> `delta_functions` and `has_delta_columns` -> `has_delta_apply`

danolivo: Hmm, we may maintain two approaches (core-based and extension-based), but they must be kept separate. So, their delta_apply structures can't intersect. That's why they have different names. If we decide to stick to the extension-based approach, one more commit might rename everything.

include/spock.h
mason-sharp: It looks like they use `pgative_` everywhere for their #defines, whereas we use `SPOCK_` everywhere for our existing #defines. It would be more consistent with our existing code to use `SPOCK_`.

danolivo: Ok, there are no objectives from my side - it's just too trivial stuff.

sql/spock--6.0.0-devel.sql
mason-sharp: It may be worth avoiding sending unneeded column data (although perhaps we want to allow doing that in the future to be able to convert UPDATEs to INSERTs, but that is a separate topic).

Re invalidation, can it be done in `spock_object_relabel`?

Otherwise, if using `REPLICA IDENTITY FULL` to trigger it, although it is a kludge, what if we immediately followed it with an `ALTER TABLE REPLICA IDENTITY DEFAULT`? Or find some benign `ALTER TABLE` variant.

(Either way we will need to revisit this once we revisit actually supporting `REPLICA IDENTITY FULL`)

danolivo: > It may be worth avoiding sending unneeded column data

Currently, we have no 'unneeded' cases. In the async LR, no one knows which value to anticipate on the other side - especially in the case of conflict. So, right now, delta is calculated each time; accordingly, the old value must be included. The current proposal seeks to change the existing conflict-resolution logic. I have no opinion on that right now - I prefer to read the initial RFC or other specs in advance.

Anyway, as a rule of thumb, I do recommend dividing complex tasks into smaller ones rather than combining them.

mason-sharp: Question about `REPLICA IDENTITY FULL`, if we set that, will it internally do a scan instead of using the unique index?

danolivo: I thought I resolved this case with the code and the comment:

```
		if (entry->has_delta_apply)
		{
			/*
			 * It looks like a hack — which, in fact, it is.
			 * We assume that delta_apply may be used for the DEFAULT identity
			 * only and will be immediately removed after altering the table.
			 * Also, if an ERROR happens here we will stay with an inconsistent
			 * value of the relreplident field. But it is just a cache ...
			 */
			relinfo->ri_RelationDesc->rd_rel->relreplident = REPLICA_IDENTITY_DEFAULT;
			entry->idxoid = RelationGetReplicaIndex(relinfo->ri_RelationDesc);
			Assert(entry->idxoid != InvalidOid);
			relinfo->ri_RelationDesc->rd_rel->relreplident = REPLICA_IDENTITY_FULL;
		}
```

Is it unclear?

mason-sharp (approved):

PR #274, opened 2025-11-27
SPOC-345: Move zodan sync test to separate github action

danolivo:
There are me also fix the issue with base image ...

### DIFF
--- a/.github/workflows/zodan_sync.yml
+++ b/.github/workflows/zodan_sync.yml
@@ -42,14 +42,7 @@ jobs:
 
       - name: Build docker container
         run: |
-          cd ${GITHUB_WORKSPACE}/
-          ls -la
-          ls -la sql
-          ls -la tests/
-          ls -la tests/tap
-          ls -la tests/tap/t
-          cat tests/tap/t/011_zodan_sync_third.pl
-          docker build \
+            docker build \
             --build-arg PGVER=${{ matrix.pgver }} \
             -t spock -f tests/docker/Dockerfile-step-1.el9 .
 
--- a/tests/docker/Dockerfile-step-1.el9
+++ b/tests/docker/Dockerfile-step-1.el9
@@ -47,18 +47,8 @@ RUN . /home/pgedge/.bashrc && export PG_CONFIG=/home/pgedge/pgedge/pg$PGVER/bin/
 RUN echo "==========Built Spock=========="
 
 #-----------------------------------------
-COPY \
-	tests/docker/entrypoint.sh \
-	tests/docker/run-tests.sh \
-	tests/docker/run-spock-regress.sh \
-	tests/docker/run-spock-tap.sh \
-	/home/pgedge
-
-RUN sudo chmod +x \
-	/home/pgedge/entrypoint.sh \
-	/home/pgedge/run-tests.sh \
-	/home/pgedge/run-spock-regress.sh \
-	/home/pgedge/run-spock-tap.sh
+COPY tests/docker/*.sh /home/pgedge/
+RUN sudo chmod +x /home/pgedge/*.sh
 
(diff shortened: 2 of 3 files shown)


danolivo:
@pct960,  doing this, I found an issue with the base image creation Dockerfile. If you can, please take a look - just to understand how worse it is and if my fix is correct.

.github/workflows/zodan_sync.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Set up docker compose
+        # Codacy wants us to use full commit SHA. This is for v1
+        uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746 # v1
+        with:
+          version: latest
+
+      - name: Build docker container
+        run: |
+          cd ${GITHUB_WORKSPACE}/
+          ls -la
```
pct960: We could remove the ls commands if the action works and debugging is done.

.github/workflows/zodan_sync.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Set up docker compose
+        # Codacy wants us to use full commit SHA. This is for v1
+        uses: docker/setup-compose-action@364cc21a5de5b1ee4a7f5f9d3fa374ce0ccde746 # v1
+        with:
+          version: latest
+
+      - name: Build docker container
+        run: |
+          cd ${GITHUB_WORKSPACE}/
```
pct960: Doesn't the action already start inside `${GITHUB_WORKSPACE}`?

tests/docker/run-spock-tap.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+###
+### Run selected TAP tests iteratively
+###
+
+
+export PG_CONFIG=/home/pgedge/pgedge/pg${PGVER}/bin/pg_config
+export PATH=/home/pgedge/pgedge/pg${PGVER}/bin:$PATH
+export LD_LIBRARY_PATH=/home/pgedge/pgedge/pg${PGVER}/lib/:$LD_LIBRARY_PATH
+export PROVE_TESTS=$proven_tests
```
pct960: `proven_tests` is defined later on, so what's the `PROVE_TESTS` variable getting here?

tests/docker/Dockerfile-step-1.el9
```diff
@@ -46,9 +47,18 @@ RUN . /home/pgedge/.bashrc && export PG_CONFIG=/home/pgedge/pgedge/pg$PGVER/bin/
 RUN echo "==========Built Spock=========="
 
 #-----------------------------------------
-COPY tests/docker/entrypoint.sh tests/docker/run-tests.sh tests/docker/run-spock-regress.sh /home/pgedge
-
-RUN sudo chmod +x /home/pgedge/entrypoint.sh /home/pgedge/run-tests.sh /home/pgedge/run-spock-regress.sh
+COPY \
```
pct960: Can be simplified to:
```
COPY tests/docker/*.sh /home/pgedge/
RUN sudo chmod +x /home/pgedge/*.sh
```

pct960 (changes requested):
Mostly looks good. Only some minor changes are needed.
I couldn't get around to running the workflow, however.
Will do it after you merge.

.github/workflows/zodan_sync.yml
danolivo: Yes, good catch.

.github/workflows/zodan_sync.yml
danolivo: @pct960 , all your changes applied

mason-sharp (approved):

PR #275, opened 2025-11-28
Improve sync event

danolivo:
I was a little irritated to get stuck in some corner cases using the sync_event. So, discovered the whole code and attempted to improve it a little.

I wonder if some sentences from the commit message may be used by @susan-pgedge to make the docs more straightforward.

### DIFF
--- a/sql/spock--6.0.0-devel.sql
+++ b/sql/spock--6.0.0-devel.sql
@@ -381,16 +381,28 @@ AS 'spock','spock_wait_slot_confirm_lsn'
 LANGUAGE C;
 
 CREATE FUNCTION spock.sub_wait_for_sync(subscription_name name)
-RETURNS void RETURNS NULL ON NULL INPUT VOLATILE LANGUAGE c AS 'MODULE_PATHNAME', 'spock_wait_for_subscription_sync_complete';
+RETURNS void RETURNS NULL ON NULL INPUT
+AS 'MODULE_PATHNAME', 'spock_wait_for_subscription_sync_complete'
+LANGUAGE C VOLATILE;
 
-CREATE FUNCTION spock.table_wait_for_sync(subscription_name name, relation regclass)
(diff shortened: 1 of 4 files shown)


ibrarahmad (changes requested):
The reorganization of variable declarations in the output plugin removes a
defensive NULL check that prevents crashes:

```c
// Current code:
case SPOCK_SYNC_EVENT_MSG:
{
    MemoryContext oldctx;
    SpockOutputData* data = (SpockOutputData*) ctx->output_plugin_private;
    TransactionId xid = InvalidTransactionId;

    if (txn != NULL)
        xid = txn->xid;
    
    oldctx = MemoryContextSwitchTo(data->context);
    
    OutputPluginPrepareWrite(ctx, true);
    spock_write_message(ctx->out,
                        xid,           // Uses InvalidTransactionId if NULL
                        message_lsn,
                        true,
                        prefix,
                        message_size,
                        message);
}
```

The proposed change replaces the NULL check with an Assert and directly
accesses the transaction ID:

```c
// Proposed change:
case SPOCK_SYNC_EVENT_MSG:
{
    MemoryContext oldctx;
    SpockOutputData *data;

    Assert(txn != NULL);
    
    data = (SpockOutputData*) ctx->output_plugin_private;
    oldctx = MemoryContextSwitchTo(data->context);
    
    OutputPluginPrepareWrite(ctx, true);
    spock_write_message(ctx->out,
                        txn->xid,      // Direct dereference without NULL check
                        message_lsn,
                        true,
                        prefix,
                        message_size,
                        message);
}
```

This change assumes txn can never be NULL for sync event messages. The
Assert only triggers in debug builds, so production builds will crash with
a segmentation fault if txn is unexpectedly NULL. The original code handled
this case gracefully by passing InvalidTransactionId. If the protocol
guarantees txn is always valid for sync events, this should be documented.

PR #276, opened 2025-12-01
Add multi-protocol support for rolling upgrades for v5.0.x -> v6.0.0

rasifr:
This change implements protocol version negotiation to enable rolling upgrades from Spock v5.0.0 through v6.0.0 without breaking replication.

Protocol 4 (v5.0.0-5.0.4): remote_insert_lsn at end of COMMIT only
Protocol 5 (v6.0.0+): remote_insert_lsn at beginning of all messages
	   (v5.0.5+): Same as v6.0.0, once backported.

Key changes:
- Protocol negotiation: Min(server_max, client_max)
- Conditional message format based on negotiated version
- Backward compatible: defaults to protocol 4 for v5.0.4 and below publishers
- Version guard: requires both nodes >= v5.0.0

### DIFF
--- a/tests/tap/schedule
+++ b/tests/tap/schedule
@@ -30,3 +30,5 @@ test: 010_zodan_add_remove_python
 #        break
 #    fi
 # done
+
+# test: 014_rolling_upgrade
--- a/tests/tap/t/014_rolling_upgrade.pl
+++ b/tests/tap/t/014_rolling_upgrade.pl
@@ -1,0 +1,560 @@
+use strict;
+use warnings;
+use Test::More tests => 55;
+use File::Path qw(make_path remove_tree);
+use Cwd qw(getcwd);
+
+use lib '.';
+use lib 't';
+use SpockTest qw(
+    create_cluster
+    destroy_cluster
+    system_or_bail
+    system_maybe
+    get_test_config
+    cross_wire
+    psql_or_bail
+    scalar_query
+);
+
+# =============================================================================
+# Configuration - modify these as needed
+# =============================================================================
+my $TEMP_BASE = "/tmp/spock_rolling_upgrade_test";
+my $V5_INSTALL = "$TEMP_BASE/v5_install";
+my $V6_INSTALL = "$TEMP_BASE/v6_install";
+my $V5_BRANCH = "origin/v5_STABLE";
+my $V6_BRANCH = "HEAD";
+
+# =============================================================================
+# Test: Rolling upgrade multi-protocol support between Spock v5.0.x and v6.0.0
+# =============================================================================
+# This test verifies that Spock supports rolling upgrades by:
+# 1. Building and installing Spock v5_STABLE to a temp location
+# 2. Building and installing Spock v6.0.0 (current branch) to another temp location
+# 3. Creating a 2-node cluster using v5 libraries
+# 4. Verifying replication works with both nodes on v5
+# 5. Upgrading node 1 to v6.0.0 while node 2 stays on v5
+# 6. Verifying bidirectional replication during mixed-version state
+# 7. Upgrading node 2 to v6.0.0
+# 8. Verifying replication works with both nodes on v6
+#
+# This tests the multi-protocol negotiation (Protocol 4 vs Protocol 5)
+# =============================================================================
+
+# Determine SPOCK_REPO path
+# Priority: 1) /home/pgedge/spock (CI and standard install), 2) detect from cwd
+my $SPOCK_REPO;
+if (-d "/home/pgedge/spock" && -f "/home/pgedge/spock/Makefile") {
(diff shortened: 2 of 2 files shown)


src/spock_proto_native.c
```diff
@@ -618,6 +699,16 @@ spock_read_commit(StringInfo in,
...
+
+	/*
+	 * Protocol version 4 includes remote_insert_lsn at the end of COMMIT
+	 * messages. Protocol version 5+ includes it at the beginning of ALL
+	 * messages (handled in apply_work).
+	 */
+	if (spock_apply_get_proto_version() == 4)
+		*remote_insert_lsn = pq_getmsgint64(in);
+	else
+		*remote_insert_lsn = InvalidXLogRecPtr;
```
mason-sharp, line 726: I see the caller needs to check for InvalidXLogRecPtr here. Can you add a comment about that in the function header?

ibrarahmad (changes requested):
The protocol negotiation logic successfully establishes version compatibility
between publisher and subscriber nodes. The approach of using the minimum of
server and client protocol versions provides proper backward compatibility
during rolling upgrades. The version guard preventing protocol 4 usage on
nodes below version 5.0.0 prevents protocol mismatch issues.

A issue exists in the commit message writing path. The protocol 5
conditional correctly prepends the remote insert LSN at the beginning, but
an unconditional write operation immediately follows:

```c
void
spock_write_commit(StringInfo out, SpockOutputData *data,
                   ReorderBufferTXN *txn, XLogRecPtr commit_lsn)
{
    uint8 flags = 0;
    XLogRecPtr remote_insert_lsn = GetXLogWriteRecPtr();

    /* Protocol version 5+ includes remote_insert_lsn at the beginning */
    if (spock_get_proto_version() >= 5)
        pq_sendint64(out, remote_insert_lsn);

    pq_sendint64(out, GetXLogWriteRecPtr());  // <-- DUPLICATE: Remove this line
    pq_sendbyte(out, 'C');
    
    // ... rest of function
    
    /* Protocol version 4 includes remote_insert_lsn at the end */
    if (spock_get_proto_version() == 4)
        pq_sendint64(out, remote_insert_lsn);
}
```

This results in duplicate LSN values being sent for protocol 5 connections.
The unconditional write should be removed as it was needed only for the
legacy protocol 4 format where the LSN appears at the message end.


The static variable approach for tracking negotiated protocol versions
introduces a concurrency hazard:

```c
static uint32 spock_negotiated_proto_version = SPOCK_PROTO_VERSION_NUM;
static uint32 spock_apply_proto_version = SPOCK_PROTO_MIN_VERSION_NUM;

void
spock_set_proto_version(uint32 version)
{
    spock_negotiated_proto_version = version;  // <-- RACE CONDITION
}
```

Multiple logical replication slots operating simultaneously with different
protocol versions will overwrite each other's state. The protocol version
should be stored in the per-connection output plugin data structure:

```c
// Recommended approach:
typedef struct SpockOutputData
{
    // ... existing fields ...
    uint32 negotiated_proto_version;  // <-- Already added, use this!
} SpockOutputData;

// Then pass 'data' parameter to functions instead of using globals:
uint32 spock_get_proto_version(SpockOutputData *data)
{
    return data->negotiated_proto_version;
}
```

This ensures each replication stream maintains its own negotiated protocol
version independently.

The commit reading function correctly handles the protocol difference where
version 4 reads the remote insert LSN from the message end while version 5
obtains it from the message prefix:

```c
void
spock_read_commit(StringInfo in,
                  XLogRecPtr *commit_lsn,
                  XLogRecPtr *end_lsn,
                  TimestampTz *committime,
                  XLogRecPtr *remote_insert_lsn)
{
    // ... read commit data ...
    
    /*
     * Protocol version 4 includes remote_insert_lsn at the end of COMMIT
     * messages. Protocol version 5+ includes it at the beginning of ALL
     * messages (handled in apply_work).
     * 
     * CALLERS: Must check (*remote_insert_lsn != InvalidXLogRecPtr)  <-- ADD THIS
     *          before using the returned value.
     */
    if (spock_apply_get_proto_version() == 4)
        *remote_insert_lsn = pq_getmsgint64(in);
    else
        *remote_insert_lsn = InvalidXLogRecPtr;
}
```

The caller responsibility for checking the invalid LSN sentinel value is
mentioned in code comments but should be emphasized in the function header
documentation to prevent misuse.

The regression test updates appropriately adjust protocol version parameters
from 3-4 to 4-5 ranges, reflecting the new minimum supported version:

```sql
-- Before:
'min_proto_version', '3',
'max_proto_version', '4',

-- After:
'min_proto_version', '4',  -- Correct: matches SPOCK_PROTO_MIN_VERSION_NUM
'max_proto_version', '5',  -- Correct: matches SPOCK_PROTO_VERSION_NUM
```

These changes ensure tests validate the new protocol negotiation behavior
correctly.

PR #279, opened 2025-12-03
Fix WAL recovery crash after apply worker failure

rasifr:
When an apply worker crashes and triggers server restart, WAL recovery
  fails to get the proper hashtable pointers in spock_rmgr_redo(). The startup process, checkpointer, and background
  workers need to re-attach to shared memory structures before accessing them.

  - Refactor initialization: create spock_shmem_init_internal() to centralize ShmemInitStruct/ShmemInitHash calls for SpockCtx, SpockHash, SpockGroupHash, and exception_log_ptr
  - Add spock_shmem_attach() for processes to attach to existing structures
  - Call spock_shmem_attach() in:
    - Startup process via spock_rmgr_startup()
    - Checkpointer via spock_checkpoint_hook()
    - Background workers via spock_worker_attach()
  - Add defensive NULL checks in spock_group_progress_update() and spock_group_resource_dump()
  - Fix assertions to handle partial initialization of LSN fields

### DIFF
--- a/include/spock_group.h
+++ b/include/spock_group.h
@@ -30,6 +30,7 @@ extern HTAB *SpockGroupHash;
 #define SPOCK_RES_DIRNAME   "spock"
 #define SPOCK_RES_DUMPFILE  "resource.dat"
 #define SPOCK_RES_TMPNAME   SPOCK_RES_DUMPFILE ".tmp"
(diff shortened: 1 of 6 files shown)


ibrarahmad (approved):
LGTM

PR #280, opened 2025-12-04
Add comprehensive defensive checks to prevent crashes.

ibrarahmad:
Hash table corruption checks in the relation metadata cache now emit warnings instead of aborting the entire postmaster. Hash removal operations that fail for entries obtained from sequential scan log descriptive warnings and terminate backends gracefully rather than forcing postmaster panics. This prevents cascade failures when dynahash bookkeeping becomes inconsistent, allowing other backends and workers to continue operating normally.

Hash table access functions verify shared memory initialization before performing operations. The group attach function validates SpockGroupHash exists before hash insertions and reports errors if hash_search fails to allocate entries. Lookup and foreach operations check for NULL hash pointers and handle gracefully by returning NULL or skipping iteration respectively. Progress update operations ensure both SpockCtx and SpockGroupHash are initialized before accessing locks or performing hash operations.

Worker progress retrieval no longer uses Assert statements that crash in production builds. The function returns NULL gracefully when called from contexts where the worker is not attached to a group, preventing crashes during partial initialization or error recovery paths. Stats reset operations check hash_search HASH_REMOVE return values and emit warnings when removal fails, preventing silent corruption propagation when dynahash internal bookkeeping becomes inconsistent.

The apply worker detach operation verifies the worker structure is properly initialized before dereferencing to decrement attachment counters. Entry count mismatches during resource dumps are treated as warnings rather than fatal errors when concurrent WAL replay legitimately modifies the hash table. Since WAL remains authoritative for recovery, treating transient count mismatches as fatal would cause unnecessary checkpointer crashes and prevent proper checkpoint completion.

The WAL recovery startup process explicitly attaches to shared memory structures during cleanup phase of custom resource manager processing. Without this attachment, the startup process would attempt to update group progress using invalid inherited pointers from the postmaster, causing segmentation faults during crash recovery. The group progress monitoring function now verifies all required hash tables are initialized before beginning iteration, preventing null pointer dereferences when querying replication status.

### DIFF
--- a/src/spock_functions.c
+++ b/src/spock_functions.c
@@ -2906,7 +2906,8 @@ Datum reset_channel_stats(PG_FUNCTION_ARGS)
 	hash_seq_init(&hash_seq, SpockHash);
 	while ((entry = hash_seq_search(&hash_seq)) != NULL)
 	{
-		hash_search(SpockHash, &entry->key, HASH_REMOVE, NULL);
+		if (hash_search(SpockHash, &entry->key, HASH_REMOVE, NULL) == NULL)
+			elog(WARNING, "spock hash table corrupted during stats reset");
 	}
 
 	LWLockRelease(SpockCtx->lock);
@@ -3324,7 +3325,7 @@ get_apply_group_progress(PG_FUNCTION_ARGS)
 	HASH_SEQ_STATUS		it;
 	SpockGroupEntry	   *e;
 
-	if (!SpockCtx || !SpockHash)
+	if (!SpockCtx || !SpockHash || !SpockGroupHash)
 		ereport(ERROR,
 				(errcode(ERRCODE_OBJECT_NOT_IN_PREREQUISITE_STATE),
 				 errmsg("spock must be loaded via shared_preload_libraries")));
--- a/src/spock_group.c
+++ b/src/spock_group.c
@@ -187,7 +187,12 @@ spock_group_attach(Oid dbid, Oid node_id, Oid remote_node_id)
 	SpockGroupEntry *e;
 	bool		found;
 
+	if (!SpockGroupHash)
+		elog(ERROR, "spock_group_attach: SpockGroupHash not initialized");
+
 	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
+	if (!e)
+		elog(ERROR, "spock_group_attach: hash_search failed");
 	if (!found)
 	{
 		/* initialize key values; Other entries will be updated later */
@@ -211,6 +216,14 @@ spock_group_attach(Oid dbid, Oid node_id, Oid remote_node_id)
 void
 spock_group_detach(void)
 {
+	/*
+	 * Only apply workers ever attach to a group. Be defensive so that a
+	 * caller in some other context (or a partially-initialized worker)
+	 * does not dereference a NULL MyApplyWorker.
+	 */
+	if (MyApplyWorker == NULL)
+		return;
+
 	if (MyApplyWorker->apply_group)
 		pg_atomic_sub_fetch_u32(&MyApplyWorker->apply_group->nattached, 1);
 
@@ -297,14 +310,25 @@ spock_group_progress_update(const SpockApplyProgress *sap)
 	if (!sap)
 		return false;
 
-	if (!SpockGroupHash || !SpockCtx)
+	/*
+	 * Safety check: WAL redo and other callers must attach to shared memory
+	 * before updating the progress hash.
+	 */
+	if (!SpockCtx || !SpockGroupHash)
 	{
-		elog(WARNING, "spock_group_progress_update: SpockGroupHash not initialized");
+		elog(WARNING,
+			 "spock_group_progress_update: shared memory not initialized");
 		return false;
 	}
 
 	key = make_key(sap->key.dbid, sap->key.node_id, sap->key.remote_node_id);
 	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
+
+	if (!e)
+	{
+		elog(ERROR, "spock_group_progress_update: hash_search failed");
+		return false;
+	}
 
 	if (!found)					/* New Entry */
 	{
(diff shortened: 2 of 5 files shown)


src/spock_functions.c
```diff
@@ -3324,7 +3325,7 @@ get_apply_group_progress(PG_FUNCTION_ARGS)
 	HASH_SEQ_STATUS		it;
 	SpockGroupEntry	   *e;
 
-	if (!SpockCtx || !SpockHash)
+	if (!SpockCtx || !SpockHash || !SpockGroupHash)
```
mason-sharp, line 3328: There is a separate check for !SpockGroupHash right below. We should use one or the other.

src/spock_group.c
```diff
@@ -297,15 +310,26 @@ spock_group_progress_update(const SpockApplyProgress *sap)
...
 		return false;
 	}
 
 	key = make_key(sap->key.dbid, sap->key.node_id, sap->key.remote_node_id);
 	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
 
+	if (!e)
+	{
+		elog(ERROR, "spock_group_progress_update: hash_search failed");
+		return false;
```
mason-sharp, line 330: Just confirming you meant `ERROR` (because of the `return`).

src/spock_functions.c
```diff
@@ -2906,7 +2906,8 @@ Datum reset_channel_stats(PG_FUNCTION_ARGS)
 	hash_seq_init(&hash_seq, SpockHash);
 	while ((entry = hash_seq_search(&hash_seq)) != NULL)
 	{
-		hash_search(SpockHash, &entry->key, HASH_REMOVE, NULL);
+		if (hash_search(SpockHash, &entry->key, HASH_REMOVE, NULL) == NULL)
+			elog(WARNING, "spock hash table corrupted during stats reset");
```
mason-sharp, line 2910: `WARNING` and "corrupted" are used together in a few places.  Should any of them be an actual ERROR?

src/spock_group.c
ibrarahmad: The return statement is used to suppress compiler warnings.

src/spock_functions.c
ibrarahmad: Yes we can have an ERROR, if we agree.

mason-sharp: Per our discussion, ERROR

src/spock_group.c
```diff
@@ -327,6 +351,13 @@ void
 spock_group_progress_update_ptr(SpockGroupEntry *e, const SpockApplyProgress *sap)
 {
 	Assert(e && sap);
+
+	if (!SpockCtx)
+	{
+		elog(WARNING, "spock_group_progress_update_ptr: SpockCtx not initialized");
```
mason-sharp, line 357: Thinking about it still.. perhaps PANIC if we keep this check.

src/spock_functions.c
rasifr: ERROR or WARNING does not make sense here. Its emptying the hash table anyway. If there was any corruption, that would have occured before this step in the function.

src/spock_group.c
```diff
@@ -187,7 +187,12 @@ spock_group_attach(Oid dbid, Oid node_id, Oid remote_node_id)
...
 	bool		found;
 
+	if (!SpockGroupHash)
+		elog(ERROR, "spock_group_attach: SpockGroupHash not initialized");
+
 	e = (SpockGroupEntry *) hash_search(SpockGroupHash, &key, HASH_ENTER, &found);
+	if (!e)
+		elog(ERROR, "spock_group_attach: hash_search failed");
 	if (!found)
 	{
```
rasifr, line 197: there is no need for this 2nd check. !found is covering it when a new entry is added to the hash. also when existing entry is found, that is returned. hash_search errors out if there is any issues, it would not return NULL for HASH_ENTER.

src/spock_group.c
rasifr, line 330: same as above.

src/spock_group.c
rasifr: True. The way we handle shared memory initialization via spock_shmem_attach(), either all Shared structures are properly initialized or something is terribly wrong.

Perhaps Assert should be used. I think that would be better option.

src/spock_group.c
```diff
@@ -342,17 +373,15 @@ spock_group_progress_update_ptr(SpockGroupEntry *e, const SpockApplyProgress *sa
...
-    Assert(MyApplyWorker != NULL);
-    Assert(MyApplyWorker->apply_group != NULL);
+	if (MyApplyWorker && MyApplyWorker->apply_group)
+		return &MyApplyWorker->apply_group->progress;
 
-    if (MyApplyWorker && MyApplyWorker->apply_group)
-        return &MyApplyWorker->apply_group->progress;
-
-    return NULL;
+	return NULL;
```
rasifr, line 384: Why remove asserts? Also not sure what this changing?

src/spock_group.c
```diff
@@ -387,6 +419,13 @@ spock_group_foreach(SpockGroupIterCB cb, void *arg)
 	SpockGroupEntry *e;
 
 	Assert(cb);
+
+	if (!SpockGroupHash)
+	{
+		elog(WARNING, "spock_group_foreach: SpockGroupHash not initialized");
+		return;
+	}
+
```
rasifr, line 428: Perhaps Assert or PANIC? I think Assert is better option.

src/spock_group.c
```diff
@@ -427,13 +466,14 @@ spock_group_resource_dump(void)
...
+	 * Safety check: if shared memory is not initialized, we cannot dump any
+	 * state.  This should not normally happen, but it is better to skip the
+	 * dump than risk dereferencing invalid pointers.
 	 */
 	if (!SpockCtx || !SpockGroupHash)
 	{
-		elog(WARNING, "spock_group_resource_dump: shared memory not initialized, skipping dump");
+		elog(WARNING,
+			 "spock_group_resource_dump: shared memory not initialized, skipping dump");
 		return;
```
rasifr, line 477: formatting issues?

PR #283, opened 2025-12-08
Move check_simple_rowfilter_expr to the spock_compat module (part of SPOC-229).

danolivo:
Sometimes, we need to call a core function that is still declared as 'static'. Before, we used a patch that did trivial, dumb stuff, making it exported one.

Here, we want to follow a more conventional approach (see, for example, the pg_hint_plan extension), where we copy the necessary function and maintain it alongside the extension. Yes, it adds some maintenance overhead, but as a huge benefit, the extension may be used with the core Postgres.

This commit removes the core patch 'pg18-035-row-filter-check'.


mason-sharp (approved):
Agreed; it is worth the maintenance headache.

PR #305, opened 2026-01-07
Fix pgindent issues and run check in Github Actions workflow

danolivo:
### Cross-version typedef solution

Different PostgreSQL major versions define different structs and enums, causing
pgindent to produce different formatting depending on which PG version's typedefs
are used. This PR solves the problem by fetching typedef lists from the PostgreSQL
buildfarm for all supported branches (PG 16, 17, 18), merging them with
Spock-specific typedefs extracted from the source tree, and deduplicating into a
single `typedefs.list`. PG 15 is excluded because `pg_bsd_indent` is not present
in its source tree and its typedefs are a subset of PG 16's.

### Changes

- **run-pgindent.sh**: Rewritten to fetch typedefs from multiple PG branches,
  clean up artifacts (empty lines, comment fragments), and support `--check`
  mode for CI dry-run
- **CI integration**: Removed the standalone `pgindent-check.yml` workflow.
  The formatting check now runs inside the existing `spockbench.yml` workflow
  after the docker build step, reusing the PG source tree already in the
  container. Skipped for the PG 15 matrix entry
- **typedefs.list**: Regenerated from PG 16-18 buildfarm lists merged with
  Spock source typedefs
- **Code consistency**: Name anonymous enums (`SpockResolveOption`,
  `GroupProgressTupDescColumns`), change `spock_conflict_resolver` from `int`
  to `SpockResolveOption`, remove duplicate extern declarations
- **pgindent run**: Applied formatting to all `.c` and `.h` files using the
  unified typedef list
- **Script cleanup**: Renamed `gen-typedefs.sh` to `run-pgindent.sh`, fixed
  duplicate typedef generation, updated README with cross-version documentation


include/spock_conflict.h
```diff
@@ -37,7 +37,7 @@ typedef enum
 	SPOCK_RESOLVE_KEEP_LOCAL,
 	SPOCK_RESOLVE_LAST_UPDATE_WINS,
 	SPOCK_RESOLVE_FIRST_UPDATE_WINS
-} SpockResolveOption;
+}			SpockResolveOption;
```
rasifr, line 41: It looks like typedefs.list is missing some names and that has disturbed the indentation.

include/spock_group.h
```diff
@@ -81,7 +81,7 @@ typedef enum
 
 	/* The last value */
 	_GP_LAST_
-} GroupProgressTupDescColumns;
+}			GroupProgressTupDescColumns;
```
rasifr: similar case

rasifr (commented):
Even though it’s a standalone CLI utility, Its still part of Spock repo and I would prefer to keep processing it, so the formatting stays consistent across the codebase.

include/spock_group.h
danolivo: @rasifr , thanks, done!

src/spock_group.c
```diff
@@ -343,12 +343,12 @@ spock_group_progress_update(const SpockApplyProgress *sap)
 		/*
 		 * This should never happen in normal operation. The shared memory
 		 * structures are initialized during postmaster startup via
-		 * shmem_startup_hook. If we hit this, it likely indicates:
-		 * 1. A bug in initialization ordering, or
-		 * 2. Corruption of shared memory pointers
+		 * shmem_startup_hook. If we hit this, it likely indicates: 1. A bug
```
mason-sharp: For comments with intentional breaks , can you use /*----- to preserve for readability?
Here we should also indent within the comment though too.

mason-sharp (commented):
I cited an example, but I think there are places where it is better to preserve the breaks or indentations.

src/spock_group.c
danolivo: @mason-sharp ,
I employed Claude to analyse our pgindent run - it found 8 comments where such 'shielding' makes sense. Done as one more commit.

mason-sharp (approved):

PR #353, opened 2026-02-18
Arbitrary improvements

danolivo:
Removed all the Windows code as we don't intend to support it.


mason-sharp (commented):
These look good to me:
- f03cef4 build_exclude_extension
- 2ef966a Remove Windows
- 4babc2a spock_disable_description

Re: check_all_uc_indexes, it is used if it can’t find the replica identity to try to find a fallback key to use, different than the description. I am not sure if we should have a comment for it in the enum definition anyway.

Regarding the formatting, I don’t think I would want us to modify the previously released sql files. For spock—6.0.0-devel.sql file, that is a different story.

danolivo:
> Re: check_all_uc_indexes, it is used if it can’t find the replica identity to try to find a fallback key to use, different than the description. I am not sure if we should have a comment for it in the enum definition anyway.

I think we must: by this, we break vanilla's behaviour (and ignore the same in UPDATE as well). So, we should say words on that. I just spent hours, if not a day, realising why my code and tests don't work because I rely on my knowledge of upstream Postgres. So, to save time and let people understand why we go against the upstream, it makes sense.

> 
> Regarding the formatting, I don’t think I would want us to modify the previously released sql files. For spock—6.0.0-devel.sql file, that is a different story.

Ok. I just note that formatting doesn't break anything, even in the old versions.

include/spock_conflict.h
```diff
@@ -51,7 +51,16 @@ extern int	spock_resolutions_retention_days;
...
-	/* The row to be inserted violates unique constraint */
+	/*
+	 * The row to be inserted violates a unique constraint.
+	 * This behaviour is controlled by GUC check_all_uc_indexes (default OFF).
+	 * When ON, this conflict exactly matches PostgreSQL's INSERT_EXISTS
+	 * conflict and Spock attempts to resolve it when any UNIQUE index rejects
+	 * the insertion.  It is counted as a conflict in subscription statistics
+	 * and the exception log.  When OFF, only a replica identity index
+	 * violation is counted as a conflict; other index violations cause an
+	 * ERROR and are not counted as conflicts.
```
mason-sharp, line 62: When `check_all_uc_indexes` is on, a UC index match is only consulted as a _fallback_ key when the RI lookup misses.

I think it should be something like this:
  /*
   * The row to be inserted violates a unique constraint.
   * When the GUC check_all_uc_indexes is ON (default OFF) and the
   * replica identity index does not find an existing tuple, Spock
   * falls back to checking all other unique indexes on the relation.
   * If any of them find a match, the insertion is treated as this
   * conflict type, counted in subscription statistics and the
   * exception log, and Spock attempts to resolve it. If the
   * incoming row is determined to be newer with last_update_wins, the
   * INSERT is converted to an UPDATE of the existing row.  When OFF,
   * only a replica identity index violation is counted as a conflict;
   * other unique index violations cause an ERROR as usual.
   */

mason-sharp:
> > Re: check_all_uc_indexes, it is used if it can’t find the replica identity to try to find a fallback key to use, different than the description. I am not sure if we should have a comment for it in the enum definition anyway.
> 
> I think we must: by this, we break vanilla's behaviour (and ignore the same in UPDATE as well). So, we should say words on that. I just spent hours, if not a day, realising why my code and tests don't work because I rely on my knowledge of upstream Postgres. So, to save time and let people understand why we go against the upstream, it makes sense.

By that I meant that the comments that were added are not accurate, I proposed some other wording just now.

> 
> > Regarding the formatting, I don’t think I would want us to modify the previously released sql files. For spock—6.0.0-devel.sql file, that is a different story.
> 
> Ok. I just note that formatting doesn't break anything, even in the old versions.

Yes, I know. I prefer to leave it in its state at the time of the release, avoiding any git history changes after.

Feel free to break up this PR into multiple ones to get changes in more easily, like the windows-related one.

include/spock_conflict.h
danolivo: Let me disagree with your variant.

It describes *what* code does - I can easily catch it from the code itself. But my point is to identify where Spock strays away from the mainstream approach to cause discussion *why* do we need it. This specific behaviour opens a hole for consistency breaches (we have discussed it and even written it down), and it would be nice to understand why, at least.

danolivo:
> I prefer to leave

That's exactly what I try to understand - what is the actual reasoning for such a 'preference'? We have automation tools to verify the equivalence between the previous and new SQL scripts. Why not do it properly?

mason-sharp:
Closing. Let's break this out into smaller PRs

PR #373, opened 2026-03-04
Replace wal_sender_timeout-based liveness with TCP keepalive.

ibrarahmad:
The apply worker previously relied on wal_sender_timeout as both a server-side disconnect trigger and an indirect keepalive pressure on the subscriber.  This caused spurious disconnects in two scenarios: a flood of 'w' messages keeping the subscriber too busy to send 'r' feedback in time, and large transactions whose apply time exceeded wal_sender_timeout.

The workaround was maybe_send_feedback(), which force-sent 'r' after every 10 'w' messages or wal_sender_timeout/2, whichever came first. This was a fragile band-aid that coupled subscriber behavior to a server GUC it cannot control.

Replace the entire mechanism with a clean two-layer model:

- TCP keepalive (keepalives_idle=10, keepalives_interval=5, keepalives_count=3) is the primary liveness detector on both sides. A dead network or crashed host is detected in ~25 seconds.

- wal_sender_timeout=0 is set on replication connections so the walsender never disconnects due to missing 'r' feedback.  Liveness on the server side is now handled entirely by TCP keepalive.

- spock.apply_idle_timeout (default 300s) is a subscriber-side safety net for a hung-but-connected walsender whose TCP keepalive probes are answered by the kernel but sends no data.  Set to 0 to disable.

Fix a bug in last_receive_timestamp handling: it was updated unconditionally after every PQgetCopyData call, including when r==0 (no data available).  Each 1-second WL_TIMEOUT spin silently reset the timer, making apply_idle_timeout never fire.  Move the update to after the r==0 guard so it reflects actual data receipt only.

Remove maybe_send_feedback() as it is no longer needed.

SPOC-419

### DIFF
--- a/src/spock.c
+++ b/src/spock.c
@@ -368,21 +368,6 @@ spock_connect_base(const char *connstr, const char *appname,
 	keys[i] = "replication";
 	vals[i] = replication ? "database" : NULL;
 	i++;
-	/*
-	 * For replication connections, disable the server-side walsender timeout.
-	 * Liveness detection is handled by TCP keepalives (keepalives_idle /
-	 * keepalives_interval / keepalives_count above) on both sides, and by
-	 * spock.apply_idle_timeout on the subscriber side as a safety net for a
-	 * hung-but-connected walsender.  Leaving wal_sender_timeout active would
-	 * cause spurious disconnects whenever the subscriber is legitimately busy
-	 * applying a large transaction and cannot send 'r' feedback in time.
-	 */
-	if (replication)
-	{
-		keys[i] = "options";
-		vals[i] = "-c wal_sender_timeout=0";
-		i++;
-	}
 	keys[i] = NULL;
 	vals[i] = NULL;
 
--- a/src/spock.c
+++ b/src/spock.c
@@ -148,6 +148,7 @@ int			spock_feedback_frequency;
 bool		check_all_uc_indexes = false;
 bool		spock_enable_quiet_mode = false;
 int			log_origin_change = SPOCK_ORIGIN_NONE;
+int			spock_apply_idle_timeout = 300;
 
 static emit_log_hook_type prev_emit_log_hook = NULL;
 static Checkpoint_hook_type prev_Checkpoint_hook = NULL;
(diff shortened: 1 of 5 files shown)


src/spock_apply.c
```diff
@@ -2838,25 +2836,24 @@ apply_work(PGconn *streamConn)
...
-			 * Note: keepalive configuration is supposed to cover this but is
-			 * apparently unreliable.
+			 * Connection liveness is handled by TCP keepalive (primary)
+			 * and PQstatus == CONNECTION_BAD (above). The idle timeout
+			 * below is a safety net for the case where the walsender
+			 * process is alive but hung -- TCP probes succeed because the
+			 * kernel ACKs them, but no data is being sent.
 			 */
-			if (rc & WL_TIMEOUT)
+			if (rc & WL_TIMEOUT && spock_apply_idle_timeout > 0)
```
danolivo, line 2974: It seems like if walsender just doesn't have data to send for a long time, subscriber will restart. Am I wrong?

It would be better to modify walsender little: skip keepalive messages being busy and rely on TCP status. But send keepalive messages if no data arrives from the WAL. In this case we don't need any subscriber-side GUC at all.

src/spock.c
```diff
@@ -345,17 +346,32 @@ spock_connect_base(const char *connstr, const char *appname,
...
+	 * keepalives_interval / keepalives_count above) on both sides, and by
+	 * spock.apply_idle_timeout on the subscriber side as a safety net for a
+	 * hung-but-connected walsender.  Leaving wal_sender_timeout active would
+	 * cause spurious disconnects whenever the subscriber is legitimately busy
+	 * applying a large transaction and cannot send 'r' feedback in time.
+	 */
+	if (replication)
+	{
+		keys[i] = "options";
+		vals[i] = "-c wal_sender_timeout=0";
```
mason-sharp: Maybe this should be non-zero, like 300000.  Less?

What if there is a proxy that blocks the sender when it is trying to send WAL? The keepalives sent from the idle receiver won't make a difference.  Meanwhile, WAL is accumulating.

mason-sharp (commented):
Added a comment about having a non-zero wal_sender_timeout.

Also, needs a rebase.

Also, could use that test file.

src/spock_apply.c
ibrarahmad: It does a slightly different task. We are relying on TCP_KEEPALIVE. The idle time is just for the guard; if the work is stuck at the application level, but the kernel TCP keep-alive will continue, there will be no way to restart the wal_sender.

mason-sharp (approved):

PR #376, opened 2026-03-04
Rework read-only mode: use Postgres native DefaultXactReadOnly

danolivo:
- Replace manual command-type filtering in `post_parse_analyze` and
  `ExecutorStart` hooks with setting `XactReadOnly` directly for non-superuser
  sessions, leveraging Postgres's built-in read-only transaction enforcement
- Distinguish Spock-imposed read-only from user/GUC-imposed read-only by
  comparing `XactReadOnly` against the `transaction_read_only` GUC value
- Rename `READONLY_USER` to `READONLY_LOCAL` to better reflect that the mode
  restricts local (non-replication) connections
- In the apply worker, move `CHECK_FOR_INTERRUPTS` and config reload to the
  top of the replay loop so read-only mode changes are detected before applying
  the next transaction
- When `READONLY_ALL` is active, the apply worker sends keepalive feedback but
  skips applying new transactions
- Escalate the `READONLY_ALL` violation in `replication_handler` from `ERROR`
  to `PANIC`

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -2434,7 +2434,7 @@ replication_handler(StringInfo s)
 	char		action = pq_getmsgbyte(s);
 
 	if (spock_readonly == READONLY_ALL)
-		elog(ERROR, "SPOCK %s: cluster is in read-only mode, not performing replication",
+		elog(PANIC, "SPOCK %s: cluster is in read-only mode, not performing replication",
 			 MySubscription->name);
 
 	memset(&errcallback_arg, 0, sizeof(struct ActionErrCallbackArg));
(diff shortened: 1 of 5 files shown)


src/spock_readonly.c
```diff
@@ -62,56 +65,53 @@ spockro_terminate_active_transactions(PG_FUNCTION_ARGS)
...
 {
-	bool		command_is_ro = false;
+	/*
+	 * Let's do the same job as at parse analysis hook.
+	 *
+	 * In some cases parse analysis and planning may be skipped on repeated
+	 * execution (remember SPI plan for example). So, additional control makes
+	 * sense here.
+	 */
+	if (spock_readonly >= READONLY_LOCAL && !superuser())
```
rasifr, line 102: When spock.readonly = 'all', a superuser can still execute DDL/DML, Which shouldn't be the case right?

danolivo: No, the refactoring precisely clarifies the case: LOCAL means only backends are forbidden to write; ALL means LR processes too. Superusers can write at any time they want. Disallowing that is too dangerous: in case of a bug, the only way to change mode might be SIGKILL. So, this refactoring attempts to make this feature safer.

src/spock_apply.c
```diff
@@ -2434,7 +2434,7 @@ replication_handler(StringInfo s)
 	char		action = pq_getmsgbyte(s);
 
 	if (spock_readonly == READONLY_ALL)
-		elog(ERROR, "SPOCK %s: cluster is in read-only mode, not performing replication",
+		elog(PANIC, "SPOCK %s: cluster is in read-only mode, not performing replication",
 			 MySubscription->name);
```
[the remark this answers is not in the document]
mason-sharp: Makes sense, there is no corruption of data. FATAL will just restart the worker.

ibrarahmad (approved):
LGTM

PR #380, opened 2026-03-06
Add PR template

mason-sharp:
PR template to be filled out with each PR in the future.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,3 @@
+## Pull Requests
+When creating PRs, always read and follow the template at `.github/pull_request_template.md`.
+Use `--body-file .github/pull_request_template.md` with `gh pr create`, then fill in each section appropriately based on the changes.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,22 @@
+<!-- PR title format: follow Conventional Commits style, e.g.
+     feat: add new postgres minor release images
+     fix: correctly mark stale instances as unknown
+     docs: add database migration guide
+     Common types: feat, fix, docs, chore, test, refactor, build(release)
+     See: https://www.conventionalcommits.org/en/v1.0.0/#summary
+-->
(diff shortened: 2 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
+<!-- PR title format: follow Conventional Commits style, e.g.
```
danolivo, line 1: I suggest adding this template to the main `CLAUDE.md` file, forcing Claude to use it for PR descriptions whenever needed.

mason-sharp: Good idea, I added

.claude/CLAUDE.md
```diff
@@ -0,0 +1,3 @@
+## Pull Requests
+When creating PRs, always read and follow the template at `.github/pull_request_template.md`.
+Use `--body-file .github/pull_request_template.md` with `gh pr create`, then fill in each section appropriately based on the changes.
```
danolivo, line 3: Not sure this line is necessary: I wonder how many developers are so crazy as to allow AI to execute code on their machines, but hope not so much.
As an example, you may check the same PR for snowflake extension here: https://github.com/pgEdge/snowflake/pull/37

rasifr, line 3: Addition of this file would conflict with anyone who has their own setup locally. It would be ongoing problem when creating commits or merging etc. I suggest we add a sample file (.claude/CLAUDE.md.example) instead and also have .claude/CLAUDE.md in .gitignore.

danolivo: @rasifr ,
Please, explain how it might cause problems?
As I see it, if you're using another AI, this file will be just ignored. If you use Claude and this repo, please use the standard PR format, or change this file in your personal fork. I can imagine someone wanting a personal fork, syncing it with us, and personal Claude settings for this project. It is possible, but it seems too strange and rare a case. Anyway, they have global Claude settings, and this specific one will be triggered only when a PR is created.

PR #392, opened 2026-03-16
Fix data loss when adding a node under write load.

ibrarahmad:
When a new node is added to a running cluster while the existing nodes are handling writes, the new node can end up permanently missing rows or accumulating stale updates.

Root cause: in the fallback path (Case B) of create_slot_with_progress(), the resume LSN (P_snap) was derived from ros_remote_lsn read after the slot-creation retry loop.  By that time the position could have advanced past the actual COPY boundary, creating a gap of permanently lost rows or allowing duplicate transactions to replay and corrupt balances.

Fix: capture v_pre_ros_lsn just before each slot-creation attempt and use it as the preferred P_snap fallback.  This value is guaranteed to be at or before the COPY boundary.  Add a MAX-by-LSN guard in spock_group_progress_force_set_list() to prevent overwriting a live apply worker's higher LSN with a stale P_snap value.

Also add the create_slot_with_progress() function to the upgrade SQL, update zodan.sql procedures, and adjust the TAP test timing.

### DIFF
--- a/sql/spock--6.0.0-devel.sql
+++ b/sql/spock--6.0.0-devel.sql
@@ -121,9 +121,10 @@ CREATE VIEW spock.progress AS
         SELECT oid FROM pg_database WHERE datname = current_database()
       );
 
--- Atomically create a replication slot and read spock.progress for all peers.
--- Row 0: lsn + snapshot header.  Rows 1+: one progress entry per peer.
-CREATE FUNCTION spock.create_slot_with_progress(
+-- Read peer progress (ros.remote_lsn) for all peer subscriptions.
+-- Called while apply workers are paused and the slot's snapshot is imported.
+-- Row 0: header (lsn + snapshot placeholder).  Rows 1+: one progress entry per peer.
+CREATE FUNCTION spock.read_peer_progress(
     p_slot_name text,
     p_provider_node_id oid,
     p_subscriber_node_id oid
--- a/sql/spock--6.0.0-devel.sql
+++ b/sql/spock--6.0.0-devel.sql
@@ -542,10 +542,14 @@ RETURNS void
 AS 'MODULE_PATHNAME', 'spock_pause_apply_workers'
 LANGUAGE C VOLATILE;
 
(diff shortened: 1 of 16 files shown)


src/spock_sync.c
```diff
@@ -556,6 +556,232 @@ adjust_progress_info(PGconn *origin_conn)
...
+	resetStringInfo(&query);
+
+	elog(LOG, "SPOCK cswp slot=%s provider=%u subscriber=%u",
+		 slot_name, origin_node_id, subscriber_node_id);
+
+	/*
+	 * Pause apply workers so ros.remote_lsn reflects only fully committed
+	 * state when we read it below.
+	 */
+	res = PQexec(conn, "SELECT spock.pause_apply_workers()");
```
rasifr, line 610: If an error occurs between pause and resume calls, the PG_CATCH in the caller re-throws without calling resume. Workers on the remote node will block indefinitely with no timeout. Recovery requires a manual SELECT spock.resume_apply_workers() or a server restart. Every add_node failure after the pause point leaves replication stopped on remote node until manually recovered.

sql/spock--6.0.0-devel.sql
```diff
@@ -442,6 +537,16 @@ RETURNS pg_lsn RETURNS NULL ON NULL INPUT
 AS 'MODULE_PATHNAME', 'spock_create_sync_event'
 LANGUAGE C VOLATILE;
 
+CREATE FUNCTION spock.pause_apply_workers()
+RETURNS void
+AS 'MODULE_PATHNAME', 'spock_pause_apply_workers'
+LANGUAGE C VOLATILE;
+
+CREATE FUNCTION spock.resume_apply_workers()
```
rasifr, line 547: pause_apply_workers() and resume_apply_workers() must be restricted to superusers. Any unprivileged user can call pause and then never call resume, leaving apply workers blocked indefinitely. So either add a superuser check in the C function or restrict EXECUTE in the SQL definition.

rasifr (approved):

PR #397, opened 2026-03-18
SPOC-311: Fix data loss when adding a node under write load.

ibrarahmad:
Backport the ZODAN add_node protocol fixes from main to v5.

The root cause of the data loss was that disabled subscriptions (e.g. sub_n2_n3) created by add_node started replicating from LSN 0/0 because the replication origin was never advanced.  When the apply worker received WAL from the slot's consistent point, any UPDATE on a row not yet present on the new node caused repeated "row not found" errors, which led to subscription disable and a 20-minute timeout.

Fix this by advancing both the replication slot and the replication origin to resume_lsn (the last commit LSN from spock.progress) in Phase 7 of the add_node protocol.  Also add a source-node commit catchup wait in Phase 3 to close the [resume_lsn, L_slot) gap.

Additional hardening of the ZODAN protocol:

  - Check wait_for_sync_event() return value instead of discarding it with PERFORM; raise on timeout.
  - Pass wait_if_disabled=true to wait_for_sync_event() so the procedure tolerates not-yet-enabled subscriptions.
  - Reduce all wait timeouts from 1200s to 180s and use clock_timestamp()-based bounds instead of iteration counters.
  - Detect terminal subscription states (disabled/down) and fail fast with RAISE EXCEPTION instead of silently continuing.
  - Replace RAISE NOTICE + CONTINUE with RAISE EXCEPTION throughout, so errors propagate instead of being swallowed.
  - Add wait_for_replication_catchup_with_dblink() for bounded lag-tracker polling.
  - Add bounded SQL loop replacing C-level sub_wait_for_sync() in Phase 7 to avoid rare hangs.
  - Drop stale replication origins before creating disabled subscriptions to prevent stale-LSN data loss.
  - Advance replication origin on new node alongside slot advancement.

Extend 011_zodan_sync_third TAP test with a remove/re-add cycle to exercise the full add_node protocol under pgbench write load, and enable 012_zodan_basics in the TAP schedule.  Increase the spockbench CI TAP test timeout to 45 minutes.

### DIFF
--- a/.github/workflows/spockbench.yml
+++ b/.github/workflows/spockbench.yml
@@ -85,6 +85,7 @@ jobs:
           echo "TAP_CT_NAME=$TAP_CT_NAME" >> "$GITHUB_ENV"
           docker run --name "$TAP_CT_NAME" -e PGVER=${{ matrix.pgver }} --workdir=/home/pgedge/spock/tests/tap \
             spock /home/pgedge/spock/tests/tap/run_tests.sh
+        timeout-minutes: 45
 
       - name: Collect TAP artifacts (from container)
         if: ${{ always() }}
--- a/.github/workflows/zodan_sync.yml
+++ b/.github/workflows/zodan_sync.yml
@@ -1,0 +1,63 @@
+# Keep its logic in sync with spockbench.yml!
+
+name: Z0DAN Sync long-lasting test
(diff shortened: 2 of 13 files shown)


mason-sharp:
Closing, will be reopened, or more likely redone, after PR #392

PR #398, opened 2026-03-23
Restore replay queue behavior

mason-sharp:
No longer make the queue size unlimited.

Also increase spock.exception_replay_queue_size default from 4MB to 16MB.

More improvements will be done in the future.


mason-sharp:
Closing because this probably won't be needed after PR #402

PR #399, opened 2026-03-24
Update spock_release_notes.md

susan-pgedge:
I've made a few minor formatting edits.

### DIFF
--- a/docs/spock_release_notes.md
+++ b/docs/spock_release_notes.md
@@ -3,18 +3,21 @@
 ## Spock 5.0.6
 
 ### New Features
-* New `spock.feedback_frequency` GUC that controls how often feedback is sent to the WAL sender. Feedback is sent every *n* messages, where *n* is the configured value. Note that feedback is also sent every wal_sender_feedback / 2 seconds.
+
+* New `spock.feedback_frequency` GUC that controls how often feedback is sent to the WAL sender. Feedback is sent every *n* messages, where *n* is the configured value. Note that feedback is also sent to every wal_sender_feedback / 2 seconds.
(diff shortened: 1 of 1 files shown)


docs/spock_release_notes.md
```diff
@@ -3,18 +3,21 @@
 ## Spock 5.0.6
 
 ### New Features
-* New `spock.feedback_frequency` GUC that controls how often feedback is sent to the WAL sender. Feedback is sent every *n* messages, where *n* is the configured value. Note that feedback is also sent every wal_sender_feedback / 2 seconds.
+
+* New `spock.feedback_frequency` GUC that controls how often feedback is sent to the WAL sender. Feedback is sent every *n* messages, where *n* is the configured value. Note that feedback is also sent to every wal_sender_feedback / 2 seconds.
```
danolivo, line 7: Maybe I missed something, but from a user perspective, I immediately want to ask - why? It adds extra traffic to the system, doesn't it? So, a few words explaining why it's necessary and how it benefits the user (including how to deactivate it if they don't need that feature) would be helpful.

docs/spock_release_notes.md
```diff
@@ -3,18 +3,21 @@
...
 * New `spock.log_origin_change` GUC to control logging of row origin changes to the PostgreSQL log. Origin changes caused by replication are no longer written to the `spock.resolutions` table, as they are informational and not true conflicts. Three modes are available:
```
danolivo, line 8: This comment is for the line out of the diff. So, I beg for your excuse.
Naming `log_origin_change` can be confusing for users: the actual action is an UPDATE from a node with a different origin ID than the one that previously changed this row. Therefore, this operation doesn't change the origin ID; it updates the row. A more suitable name could be `spock.log_alter_source_node` or something similar. This indicates that the UPDATE originated from a different node and that the user performing the update might not have had complete information about the current state of the table on the origin node of this row, possibly causing someone's data to be skipped during conflict resolution. If users are particularly cautious, they can enable this GUC.

PR #400, opened 2026-03-24
TRANSDISCARD/SUB_DISABLE: switch from subtransaction replay to read-only exception row capture

danolivo:
- In TRANSDISCARD and SUB_DISABLE modes, replace per-DML subtransaction replay with a single read-only transaction. Each DML handler logs a row to `spock.exception_log` instead of attempting the operation. Only the record that originally caused the error carries the real error message; other records get an empty string.
- Identify the failed record by `xact_action_counter` (saved as `failed_action` in `SpockExceptionLog`), which is unique per DML within a transaction.
- In DISCARD mode, each DML still runs inside a subtransaction; only the failed ones are logged.
- Add `exception_row_capture` regression test covering all three exception modes with different breakage methods (absent table, truncated table, deleted row) and an INSERT_EXISTS conflict to verify that dry-run modes leave `spock.resolutions` empty while DISCARD populates it.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -2264,33 +2264,17 @@ handle_sql_or_exception(QueuedMessage *queued_message, bool tx_just_started)
 			exception_behaviour == SUB_DISABLE)
 		{
 			/*
-			 * TRANSDISCARD and SUB_DISABLE: skip the DDL and log the
-			 * discarded operation directly to spock.exception_log.
+			 * TRANSDISCARD and SUB_DISABLE: skip the DDL, just extract SQL
+			 * for logging below.
 			 */
-			char	   *error_msg =
-				(my_exception_log_index >= 0 &&
-				 xact_action_counter ==
-				 exception_log_ptr[my_exception_log_index].failed_action &&
-				 exception_log_ptr[my_exception_log_index].initial_error_message[0] != '\0') ?
-				exception_log_ptr[my_exception_log_index].initial_error_message :
-				NULL;
-
 			sql = JsonbToCString(NULL,
 								 &queued_message->message->root, 0);
 			exception_command_counter++;
-			add_entry_to_exception_log(remote_origin_id,
-									   replorigin_session_origin_timestamp,
-									   remote_xid,
-									   0, 0,
-									   NULL, NULL, NULL, NULL,
-									   sql, queued_message->role,
-									   "SQL",
-									   error_msg);
 			failed = false;
 		}
 		else
 		{
-			/* DISCARD MODE needs hard way - try block and subtransactions */
+			/* DISCARD mode: try block with subtransactions */
 			PG_TRY();
 			{
 				exception_command_counter++;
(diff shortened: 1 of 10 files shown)


src/spock_exception_handler.c
```diff
@@ -188,12 +188,8 @@ add_entry_to_exception_log(Oid remote_origin, TimestampTz remote_commit_ts,
 		values[Anum_exception_log_ddl_user - 1] = CStringGetTextDatum(ddl_user);
 	}
 
-	/*
-	 * The error_message column of the spock.exception_log table is marked as
-	 * NOT NULL, but we don't always have a valid error message.
-	 */
 	if (error_message == NULL)
-		values[Anum_exception_log_error_message - 1] = CStringGetTextDatum("unknown");
+		values[Anum_exception_log_error_message - 1] = CStringGetTextDatum("");
```
mason-sharp: I try to avoid using empty strings because I have seen users get confused between empty strings and NULL values when querying, but "unknown" is also not great. Just adding a comment in case you or anyone has a better suggestion. Perhaps "unavailable"?

mason-sharp (commented):
I like this, it is much better than what we were doing before. I will revisit after the read-only PR.

src/spock_apply.c
```diff
@@ -769,8 +786,15 @@ handle_commit(StringInfo s)
...
+		 *
+		 * For SUB_DISABLE mode during a retry (use_try_block), do not advance
+		 * the LSN even if the replay succeeded. This allows the transaction
+		 * to be re-applied after the user fixes the root cause and re-enables
+		 * the subscription.
 		 */
-		if (!xact_had_exception ||
+		if ((!xact_had_exception &&
+			 !(MyApplyWorker->use_try_block &&
+			   exception_behaviour == SUB_DISABLE)) ||
```
mason-sharp, line 797: Good idea

src/spock_apply.c
```diff
@@ -1308,50 +1289,83 @@ handle_insert(StringInfo s)
...
+				 exception_log_ptr[my_exception_log_index].initial_error_message[0]) ?
+				exception_log_ptr[my_exception_log_index].initial_error_message :
+				NULL;
+
 			exception_command_counter++;
-			BeginInternalSubTransaction(NULL);
-			spock_apply_heap_insert(rel, &newtup);
+			exception_log_ptr[my_exception_log_index].local_tuple = NULL;
+			log_insert_exception(false, error_msg, rel, NULL, &newtup, "INSERT");
 		}
```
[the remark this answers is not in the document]
danolivo: This is OK - we add as much information as possible here

tests/regress/sql/dry_run_logging.sql
```diff
@@ -0,0 +1,282 @@
+--
+-- Test: Exception behaviour modes (DISCARD, TRANSDISCARD, SUB_DISABLE)
+--
+-- Common scenario: three tables on provider, one broken on subscriber.
+-- A single transaction with DMLs on all three tables triggers an error
+-- on the broken table.  The first DML (INSERT into drl_t1) also creates
+-- a conflict (INSERT_EXISTS) to verify that dry-run modes do not log
```
mason-sharp: Somewhat nit picky, but I think we should use a different term since dry-run implies "what would happen". How about we use "log-only replay"?

After changing throughout and the read-only PR gets in, we can get this merged in.

danolivo: @mason-sharp . It is up to you as a native speaker. I've never seen 'log-only replay' and used the 'dry-run mode' as a standard term in the industry. Feel free to commit your change at the top of the branch.

mason-sharp (approved):

PR #407, opened 2026-04-06
Add regression tests for read-only mode (spock.readonly GUC)

mason-sharp:
Cover all three modes (off, local, all), superuser exemption, backward-compatible 'user' alias, user-imposed transaction_read_only preservation, and ALTER SYSTEM with separate non-superuser connection.

### DIFF
--- a/tests/regress/expected/read_only.out
+++ b/tests/regress/expected/read_only.out
@@ -1,0 +1,223 @@
+-- Tests for read-only mode (spock.readonly GUC)
+-- Verifies three modes (off, local, all), superuser exemption,
+-- backward-compatible 'user' alias, and user-imposed transaction_read_only.
+SELECT * FROM spock_regress_variables()
+\gset
+\c :provider_dsn
+-- Store superuser name for reconnections after user switching
+SELECT current_user AS regress_su
+\gset
+-- Setup: test table with grants for nonsuper
+CREATE TABLE public.ro_test (id int primary key, data text);
+GRANT ALL ON public.ro_test TO nonsuper;
+-- Verify default state is 'off'
+SHOW spock.readonly;
+ spock.readonly 
+----------------
+ off
+(1 row)
+
+---
+--- Test: non-superuser can write when readonly is off
+---
+SET ROLE nonsuper;
+INSERT INTO ro_test VALUES (1, 'nonsuper writes when off');
+RESET ROLE;
+SELECT count(*) FROM ro_test;
+ count 
+-------
+     1
+(1 row)
+
+---
+--- Test: set readonly to 'local', superuser can still write
+---
+SET spock.readonly = 'local';
+SHOW spock.readonly;
+ spock.readonly 
+----------------
+ local
+(1 row)
+
+INSERT INTO ro_test VALUES (2, 'superuser writes in local');
+SELECT count(*) FROM ro_test;
+ count 
+-------
+     2
+(1 row)
+
+---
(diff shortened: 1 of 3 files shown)


tests/regress/expected/read_only.out
```diff
@@ -0,0 +1,223 @@
...
+
+-- Reset system setting
+ALTER SYSTEM RESET spock.readonly;
+SELECT pg_reload_conf();
+ pg_reload_conf 
+----------------
+ t
+(1 row)
+
+SELECT pg_sleep(0.5);
```
danolivo: Why do we need this delay?

tests/regress/expected/read_only.out
```diff
@@ -0,0 +1,223 @@
+-- Tests for read-only mode (spock.readonly GUC)
```
danolivo, line 1: It seems basic coverage looks ok. What I missed here:
1. DDL versus general UTILITY commands - show what spock limits and what allows execution (vacuum, SET, etc).
2. SECURITY DEFINER function - something like pgaudit still writes to the database. We should pin the current behaviour in a test.
3. The most important one: what about replication? DML, SQL, internal Spock commands? How does a read-only system react to the attempt to apply - does it crash, flood logs or gently wait?

tests/regress/expected/read_only.out
mason-sharp: You're right, I removed it.

tests/regress/expected/read_only.out
```diff
@@ -66,6 +67,14 @@ ERROR:  cannot execute TRUNCATE TABLE in a read-only transaction
 -- DDL blocked
 CREATE TABLE public.ro_fail (id int);
 ERROR:  cannot execute CREATE TABLE in a read-only transaction
+---
+--- Test: utility commands allowed in 'local' mode (still as nonsuper)
+--- Spock sets XactReadOnly = true and delegates enforcement entirely to
+--- Postgres.  These commands are allowed because Postgres does not call
```
danolivo: It would be better to show off the reasoning _why_ such commands are allowed (doesn't impact the schema and doesn't interfere with logical replication either). There are no problems guarding the instance against other commands, too, if necessary.

tests/regress/expected/read_only.out
```diff
@@ -66,6 +67,14 @@ ERROR:  cannot execute TRUNCATE TABLE in a read-only transaction
...
+--- PreventCommandIfReadOnly() for maintenance operations.
+---
+VACUUM ro_test;
+ANALYZE ro_test;
```
danolivo: The VACUUM ANALYZE command looks more compact than two separate calls

tests/regress/expected/read_only.out
```diff
@@ -215,6 +263,212 @@ SELECT count(*) FROM ro_test;
...
+(1 row)
+
+\c :subscriber_dsn
+SELECT count(*) FROM information_schema.tables
+    WHERE table_schema = 'public' AND table_name = 'ro_autoddl_test';
+ count 
+-------
+     0
+(1 row)
+
```
danolivo, line 352: It would be nice to reproduce the case where read-only blocks replication, if such a mode exists, just to see how it actually works and which ERROR messages bubble up.

danolivo (approved):

mason-sharp:
Thanks for approving @danolivo . I decided to do one (hopefully) last commit to address recent feedback.

PR #412, opened 2026-04-09
feat: add spock.resolutions_retention_days GUC and cleanup_resolutions()

rasifr:
## Summary

Adds `spock.resolutions_retention_days` GUC (int, default 100, min 0) to control how long rows are kept in `spock.resolutions`. The apply worker deletes rows older than the configured window at most once per day. Setting to `0` disables automatic cleanup entirely.

Also adds `spock.cleanup_resolutions()`, a superuser-only SQL function that returns the number of rows deleted, for manual invocation.

A `log_time` index is added to both the fresh-install and upgrade scripts to keep the daily `DELETE` efficient on large tables.

### DIFF
--- a/docs/spock_functions/functions/spock_cleanup_resolutions.md
+++ b/docs/spock_functions/functions/spock_cleanup_resolutions.md
@@ -4,7 +4,7 @@ spock.cleanup_resolutions()
 
 ### SYNOPSIS
 
-spock.cleanup_resolutions()
+spock.cleanup_resolutions([days integer])
 
 ### RETURNS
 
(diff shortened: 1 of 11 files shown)


src/spock_apply.c
```diff
@@ -3050,6 +3054,26 @@ apply_work(PGconn *streamConn)
...
 			}
 
+			/*
+			 * Periodically clean up old rows from spock.resolutions.  We run
+			 * at most once per day regardless of whether the worker is idle
+			 * or processing traffic.  spock_cleanup_resolutions() manages its
+			 * own transaction and error handling.
+			 */
+			if (!IsTransactionState() &&
+				spock_resolutions_retention_days > 0)
```
mason-sharp, line 3064: The docs mention it checks save_resolutions, too.

src/spock.c
```diff
@@ -1012,6 +1012,17 @@ _PG_init(void)
...
 							 NULL, NULL, NULL);
 
+	DefineCustomIntVariable("spock.resolutions_retention_days",
+							"Number of days to retain rows in spock." CATALOG_LOGTABLE " table. "
+							"Rows older than this are deleted periodically by the apply worker. "
+							"Set to 0 to disable automatic cleanup.",
+							NULL,
+							&spock_resolutions_retention_days,
+							100, 0, INT_MAX,
+							PGC_SUSET,
```
mason-sharp: Should this use PGC_SIGHUP since it is used in the apply worker?

src/spock_apply.c
```diff
@@ -218,6 +218,9 @@ static dlist_head sync_replica_lsn = DLIST_STATIC_INIT(sync_replica_lsn);
 static XLogRecPtr skip_xact_finish_lsn = InvalidXLogRecPtr;
 #define is_skipping_changes() (unlikely(!XLogRecPtrIsInvalid(skip_xact_finish_lsn)))
 
+/* How often the apply worker runs spock_cleanup_resolutions() (milliseconds). */
+#define RESOLUTIONS_CLEANUP_INTERVAL_MS (86400L * 1000L)
```
mason-sharp, line 222: We could leave this, but ms resolution seems like overkill when seconds would do?

docs/spock_functions/functions/spock_cleanup_resolutions.md
```diff
@@ -0,0 +1,41 @@
+## NAME
+
+spock.cleanup_resolutions()
```
mason-sharp, line 3: well, since you created this, what might be nice is if takes an optional argument that overrides resolutions_retention_days. For example, maybe they set retention to 0, then periodically manually run something like `SELECT spock.cleanup_resolutions(60)` when they think it is getting too big.

src/spock_apply.c
```diff
@@ -3050,6 +3054,26 @@ apply_work(PGconn *streamConn)
...
+				TimestampTz cleanup_due;
+
+				cleanup_due = TimestampTzPlusMilliseconds(last_cleanup_timestamp,
+														  RESOLUTIONS_CLEANUP_INTERVAL_MS);
+				if (GetCurrentTimestamp() >= cleanup_due)
+				{
+					spock_cleanup_resolutions();
+					last_cleanup_timestamp = GetCurrentTimestamp();
+				}
+			}
```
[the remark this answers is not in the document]
mason-sharp: @coderabbit Isn't it already wrapped in a PG_TRY-PG_CATCH in spock_cleanup_resolutions()?

mason-sharp (changes requested):

mason-sharp (approved):

PR #413, opened 2026-04-09
SPOC-437: Add `spock.sub_alter_options()` for bulk subscription option changes

danolivo:
Introduce a new SQL-callable function that accepts a subscription name and a JSONB object to modify multiple subscription settings in a single call:

```sql
spock.sub_alter_options(subscription_name name, options jsonb) RETURNS boolean
```

### Recognised keys

| Key | Type | Description |
|-----|------|-------------|
| `"forward_origins"` | `string[]` | Replace the list of origin names to forward (pass `["all"]` to forward all origins) |
| `"apply_delay"` | `string` | Interval string accepted by PostgreSQL's `interval_in`, e.g. `"500ms"` or `"1 minute"` |
| `"skip_schema"` | `string[]` | Replace the list of schemas to skip during apply |

Only keys present in the JSONB object are modified; omitted settings retain their current values. Any unrecognised key or wrong JSON type raises an error immediately.

### Why JSONB rather than separate functions

The existing `sub_alter_*` family requires one round-trip per option. When several settings need to change atomically (e.g. `forward_origins` + `apply_delay` together), callers had to issue multiple calls and the subscription catalog was updated — and the apply worker signalled — multiple times. `sub_alter_options` performs a single catalog write and sends one SIGTERM, which the manager uses to restart the apply worker with the new settings.

The JSONB interface is also forward-compatible: new options can be added without new C functions or SQL signatures.

### Restart behaviour

`alter_subscription()` sends SIGTERM to the apply worker at transaction commit; the spock manager restarts it automatically. Options that affect the replication stream (`forward_origins`, `apply_delay`) therefore take effect on the next `START_REPLICATION` negotiation without any manual `sub_disable`/`sub_enable` cycle.

### Files changed

- `src/spock_functions.c` — `spock_alter_subscription_options` C implementation + `PG_FUNCTION_INFO_V1` declaration; added `#include "utils/jsonb.h"`
- `sql/spock--6.0.0-devel.sql` — `CREATE FUNCTION` for the fresh-install path
- `sql/spock--5.0.6--6.0.0-devel.sql` — `CREATE FUNCTION` for the upgrade path
- `tests/tap/t/018_forward_origins.pl` — TAP test: 3-node cascade (n1 → n2 → n3), confirms n1 inserts do not reach n3 while forwarding is disabled, then enables forwarding via `sub_alter_options` and verifies n1 inserts now propagate end-to-end; uses `spock.sync_event` / `wait_for_sync_event` for deterministic synchronisation instead of sleeps
- `tests/tap/schedule` — registers the new test

### DIFF
--- a/src/spock_functions.c
+++ b/src/spock_functions.c
@@ -72,6 +72,7 @@
 #include "utils/fmgroids.h"
 #include "utils/inval.h"
 #include "utils/json.h"
+#include "utils/jsonb.h"
 #include "utils/guc.h"
 #if PG_VERSION_NUM >= 160000
 #include "utils/guc_hooks.h"
@@ -122,6 +123,7 @@ PG_FUNCTION_INFO_V1(spock_alter_subscription_enable);
 PG_FUNCTION_INFO_V1(spock_alter_subscription_add_replication_set);
 PG_FUNCTION_INFO_V1(spock_alter_subscription_remove_replication_set);
 PG_FUNCTION_INFO_V1(spock_alter_subscription_skip_lsn);
+PG_FUNCTION_INFO_V1(spock_alter_subscription_options);
 
 PG_FUNCTION_INFO_V1(spock_alter_subscription_synchronize);
 PG_FUNCTION_INFO_V1(spock_alter_subscription_resynchronize_table);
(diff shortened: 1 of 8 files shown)


src/spock_functions.c
```diff
@@ -924,6 +926,111 @@ spock_alter_subscription_skip_lsn(PG_FUNCTION_ARGS)
...
+			ereport(ERROR,
+					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					 errmsg("unrecognized subscription option \"%s\"", key),
+					 errhint("Valid options are: \"forward_origins\", "
+							 "\"apply_delay\", \"skip_schema\".")));
+		}
+	}
+
+	alter_subscription(sub);
+
```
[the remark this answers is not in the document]
danolivo: Done

src/spock_functions.c
```diff
@@ -924,6 +926,127 @@ spock_alter_subscription_skip_lsn(PG_FUNCTION_ARGS)
...
+/*
+ * Update multiple subscription options at once mentioned at the input JSONB.
+ *
+ * Recognised keys: "forward_origins", "apply_delay", and "skip_schema".
+ *
+ * Any unrecognised key or wrong JSON type raises an ERROR immediately.
+ *
+ * Note: options that affect the replication stream (forward_origins,
+ * apply_delay) take effect only after the apply worker reconnects to the
+ * publisher.  Call sub_disable() followed by sub_enable() to force an
```
mason-sharp: Is this needed (alter_subscription() restarts?)?

mason-sharp (commented):
Looking good to me, I just added one comment.

Other than that, can you update docs to add the new function?

src/spock_functions.c
danolivo: This comment highlights the issue that we don't have any guarantee or detection mechanism to determine whether the subscription has already been reloaded with new parameters. Having the 'generation' field in the background worker, we could identify such situations and death loops as well. But for now, only the disable/enable trick guarantees we have updated the state.

mason-sharp (approved):

PR #414, opened 2026-04-12
Fix Codacy false-positive findings and unused variable warnings

rasifr:
Replace strlen(PSWD_KEYWORD) with sizeof(PSWD_KEYWORD) - 1 since PSWD_KEYWORD is a compile-time string literal; evaluated at compile time with no runtime overhead.

Remove unused COL_LSN and COL_SNAP variables in
spock_create_slot_and_read_progress(); the read_peer_progress result columns 0 and 1 are intentionally skipped in favour of the values already obtained from the replication protocol.

### DIFF
--- a/src/spock.c
+++ b/src/spock.c
@@ -863,7 +863,7 @@ log_message_filter(ErrorData *edata)
 	if (edata->elevel == ERROR)
 	{
 		char	   *strpos;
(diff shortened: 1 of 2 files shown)


mason-sharp (approved):
The other flagged issues appear to be typical Postgres style code.

PR #417, opened 2026-04-13
Fix incorrect subscription name in Phase 9 of ZODAN add_node

mason-sharp:
create_sub_on_new_node_to_src_node generated sub\_{subscriber}\_{provider} instead of the expected sub\_{provider}\_{subscriber} convention, causing remove-node to fail when looking up subscriptions by their expected name.

Introduce spock.gen_sub_name(provider_node, subscriber_node) helper and replace all inline 'sub_' || concatenations to prevent future mismatches.

### DIFF
--- a/samples/Z0DAN/zodan.sql
+++ b/samples/Z0DAN/zodan.sql
@@ -29,6 +29,22 @@
 -- ============================================================================
 
 -- ============================================================================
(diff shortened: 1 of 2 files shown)


ibrarahmad (approved):
LGTM, Just check If we need some docs adjustment or not.

PR #419, opened 2026-04-14
Stale socket fd in apply_work() causes spurious worker restart on connection failure

rasifr:
fd = PQsocket(applyconn) was captured once before stream_replay: and
never refreshed.  When the provider connection dies, libpq's
pqDropConnection closes conn->sock (sets it to PGINVALID_SOCKET) and
marks status CONNECTION_BAD.  On re-entry at stream_replay: with
use_try_block=true, the stale fd is passed to WaitLatchOrSocket:

  epoll_ctl(EPOLL_CTL_ADD, closed_fd) -> EINVAL
  => ERROR "epoll_ctl() failed: Invalid argument"

This hits the use_try_block=true branch in PG_CATCH, producing
LOG "error during exception handling" + PG_RE_THROW.

Fix: at stream_replay:, refresh fd = PQsocket(applyconn) and return
cleanly if the connection is already dead.  The worker exits with
proc_exit(0) and the postmaster restarts it to reconnect.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -2867,7 +2867,6 @@ apply_work(PGconn *streamConn)
 	ErrorData  *edata = NULL;
 
 	applyconn = streamConn;
-	fd = PQsocket(applyconn);
 
 	/* Init the ApplyReplayContext used to replay after an exception */
 	ApplyReplayContext = AllocSetContextCreate(TopMemoryContext,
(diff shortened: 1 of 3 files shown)


src/spock_apply.c
```diff
@@ -2896,6 +2896,25 @@ apply_work(PGconn *streamConn)
...
+	 * failure: libpq's pqDropConnection closes conn->sock and sets status to
+	 * CONNECTION_BAD before the ERROR is thrown.  The old fd value is now
+	 * stale (closed, possibly reused by the OS).  Passing it to
+	 * WaitLatchOrSocket causes epoll_ctl(EINVAL) on Linux; on macOS kqueue
+	 * silently ignores the bad fd but PQconsumeInput then fires a second
+	 * "connection to other side has died" exception — both caught as
+	 * "error during exception handling".  Return instead so the worker
+	 * restarts and reconnects cleanly.
+	 */
+	fd = PQsocket(applyconn);
```
mason-sharp, line 2910: I think we should delete the similar call on line 2870

mason-sharp (changes requested):

mason-sharp (approved):

PR #421, opened 2026-04-16
Binary patchset handshake and a working 5.0.x → 6.0.0 upgrade path

danolivo:
## Summary

This branch closes two gaps that surface when a node built on the `v5_STABLE` line is pg_upgraded to 6.0.0:

1. **No binary-level handshake between the spock extension and the patched PostgreSQL server.** An unpatched server, or one patched against a different generation of the spock core patchset, could load the extension and fail subtly much later instead of failing cleanly at LOAD time.
2. **No usable `ALTER EXTENSION spock UPDATE` path from the current stable line.** The 6.0.0 extension shipped only a direct `spock--5.0.8--6.0.0.sql` shortcut, but `origin/v5_STABLE` now ships 5.0.10. A cluster pinned at 5.0.9 or 5.0.10 dead-ended on `ALTER EXTENSION spock UPDATE` with `"no update path from 5.0.10 to 6.0.0"` — and because the manager worker drives that UPDATE on connect, the worker crash-loops after the upgrade.

It also adds a TAP test that drives the real pg_upgrade path end to end, plus supporting test/doc fixes.

### Core patchset version handshake

* `patches/{15,16,17,18}/*-spock-patchset-version.diff` — add a compile-time `SPOCK_CORE_PATCHSET_VERSION` and a runtime `SpockCorePatchsetVersion` global (in `miscadmin.h` / `globals.c`) on every supported PG branch. An unpatched server fails to dynamic-link the symbol; a server patched against a different generation reports a clear mismatch.
* `src/spock.c` — in `_PG_init`, compare the runtime `SpockCorePatchsetVersion` against the `SPOCK_CORE_PATCHSET_VERSION` the extension was compiled against and `ereport()` on disagreement. The failure mode becomes a clean error at LOAD rather than a later crash.

### Stepped 5.0.x → 6.0.0 upgrade chain

* Remove the direct `spock--5.0.8--6.0.0.sql` shortcut.
* Add no-op `spock--5.0.8--5.0.9.sql` and `spock--5.0.9--5.0.10.sql` (neither stable release changed the schema), mirroring `v5_STABLE`'s own version scripts.
* Move the full 5.x → 6.0.0 migration into `spock--5.0.10--6.0.0.sql` (body unchanged from the old shortcut), applied at the final step of the chain.

`ALTER EXTENSION spock UPDATE` now walks `5.0.8 → 5.0.9 → 5.0.10 → 6.0.0` in a single command regardless of which 5.0.x the cluster is pinned at. Future stable point releases extend the chain with another no-op hop instead of needing a fresh direct-to-6.0.0 script.

### Tests and docs

* `tests/tap/t/030_pg_upgrade_5x_to_6x.pl` (new) — builds a `v5` (`origin/v5_STABLE`) and a `HEAD` variant side by side, runs the real pg_upgrade between them, and asserts: a regression DB survives with the same user relations; the per-attribute `delta_apply` reloption (`log_old_value` / `delta_apply_function`) survives byte-for-byte; and `ALTER EXTENSION spock UPDATE` drives `pg_extension.extversion` to 6.0.0. Includes CFLAGS/CPPFLAGS scrubbing and pre-generated headers so it runs under `make check_prove`.
* `tests/tap/t/018_upgrade_schema_match.pl` — retarget the upgrade-from version from 5.0.8 to 5.0.10 (the test builds the old node from `v5_STABLE`, which now ships 5.0.10; its `"starts at 5.0.8"` assertion would otherwise fail).
* `tests/tap/t/002_create_subscriber.pl` — replace the flaky `spock.sub_wait_for_sync` poll with the deterministic `sync_event` / `wait_for_sync_event` pattern.
* `docs/spock_release_notes.md` — generalise the Upgrading section: the single `ALTER EXTENSION spock UPDATE` now applies to any 5.0.8-or-later release and walks the chain automatically.

## Compatibility / upgrade impact

* **Server/extension must match.** After this change a 6.0.0 spock extension refuses to load against a server that lacks the core patchset or carries a different patchset generation. This is intentional — it converts a silent footgun into a clear LOAD-time error — but it does mean the server binary must be patched in lockstep with the extension.
* **No data-format change.** The 5.0.x → 6.0.0 migration SQL is byte-identical to the previously shipped shortcut; only the path the updater walks changed.
* **Clusters pinned at exactly 5.0.8** still upgrade — via the new stepped chain rather than the removed shortcut.

### DIFF
--- a/patches/16/pg16-000-spock-patchset-version.diff
+++ b/patches/16/pg16-000-spock-patchset-version.diff
@@ -1,0 +1,28 @@
+Spock core-patchset: export patchset version via miscadmin.h and globals.c.
+
+Adds SPOCK_CORE_PATCHSET_VERSION (compile-time constant) and
+SpockCorePatchsetVersion (runtime global) to the standard places
+PostgreSQL already uses for server-wide state.  No new files.
+
+--- a/src/include/miscadmin.h
++++ b/src/include/miscadmin.h
+@@ -510,4 +510,11 @@ extern void RestoreClientConnectionInfo(char *conninfo);
+ /* in executor/nodeHash.c */
+ extern size_t get_hash_memory_limit(void);
+
++/*
++ * Spock core-patchset identity.  Bump the version when the patchset
++ * changes in a way visible to the extension binary.
++ */
++#define SPOCK_CORE_PATCHSET_VERSION 1
++extern PGDLLIMPORT const int SpockCorePatchsetVersion;
++
+ #endif							/* MISCADMIN_H */
+--- a/src/backend/utils/init/globals.c
++++ b/src/backend/utils/init/globals.c
+@@ -155,3 +155,5 @@ int64		VacuumPageDirty = 0;
+
+ int			VacuumCostBalance = 0;	/* working state for vacuum */
+ bool		VacuumCostActive = false;
++
++const int	SpockCorePatchsetVersion = SPOCK_CORE_PATCHSET_VERSION;
--- a/patches/16/pg16-000-spock-patchset-version.diff
+++ b/patches/16/pg16-000-spock-patchset-version.diff
@@ -15,7 +15,7 @@ PostgreSQL already uses for server-wide state.  No new files.
 + * changes in a way visible to the extension binary.
 + */
 +#define SPOCK_CORE_PATCHSET_VERSION 1
-+extern PGDLLIMPORT const int SpockCorePatchsetVersion;
++extern PGDLLIMPORT int SpockCorePatchsetVersion;
 +
  #endif							/* MISCADMIN_H */
 --- a/src/backend/utils/init/globals.c
@@ -25,4 +25,4 @@ PostgreSQL already uses for server-wide state.  No new files.
  int			VacuumCostBalance = 0;	/* working state for vacuum */
  bool		VacuumCostActive = false;
 +
-+const int	SpockCorePatchsetVersion = SPOCK_CORE_PATCHSET_VERSION;
++int	SpockCorePatchsetVersion = SPOCK_CORE_PATCHSET_VERSION;
--- a/patches/16/pg16-000-spock-patchset-version.diff
+++ b/patches/16/pg16-000-spock-patchset-version.diff
@@ -1,0 +1,32 @@
+Spock core-patchset: export patchset version via miscadmin.h and globals.c.
+
+Adds SPOCK_CORE_PATCHSET_VERSION (compile-time constant) and
+SpockCorePatchsetVersion (runtime global) to the standard places
+PostgreSQL already uses for server-wide state.  No new files.
+
+--- a/src/include/miscadmin.h
++++ b/src/include/miscadmin.h
+@@ -510,4 +510,11 @@
+ /* in executor/nodeHash.c */
+ extern size_t get_hash_memory_limit(void);
+ 
++/*
++ * Spock core-patchset identity.  Bump the version when the patchset
++ * changes in a way visible to the extension binary.
++ */
++#define SPOCK_CORE_PATCHSET_VERSION 1
++extern PGDLLIMPORT int SpockCorePatchsetVersion;
++
+ #endif							/* MISCADMIN_H */
+--- a/src/backend/utils/init/globals.c
++++ b/src/backend/utils/init/globals.c
+@@ -114,6 +114,9 @@
+ bool		IsBinaryUpgrade = false;
+ bool		IsBackgroundWorker = false;
+ 
++/* Spock core-patchset identity. */
++int			SpockCorePatchsetVersion = SPOCK_CORE_PATCHSET_VERSION;
++
+ bool		ExitOnAnyError = false;
+ 
+ int			DateStyle = USE_ISO_DATES;
--- a/sql/spock--5.0.0.sql
+++ b/sql/spock--5.0.0.sql
@@ -1,0 +1,790 @@
+\echo Use "CREATE EXTENSION spock" to load this file. \quit
+
+CREATE TABLE spock.node (
+    node_id oid NOT NULL PRIMARY KEY,
+    node_name name NOT NULL UNIQUE,
+    location text,
+    country text,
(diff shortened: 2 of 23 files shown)


sql/spock--5.0.6--5.0.7.sql
```diff
@@ -0,0 +1,163 @@
...
+ * crashing on operators that validate ArrayType structure).  ANALYZE
+ * will repopulate as needed.
+ */
+DELETE FROM pg_catalog.pg_statistic
+ WHERE starelid  = 'spock.subscription'::regclass
+   AND staattnum = (
+       SELECT attnum
+       FROM   pg_catalog.pg_attribute
+       WHERE  attrelid = 'spock.subscription'::regclass
+         AND  attname  = 'sub_skip_schema');
```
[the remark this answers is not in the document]
danolivo: The allow_system_table_mods is needed when we modify the structure of a system table. It is unrelated to DML operations.

docs/internals-doc/binary-upgrade-compat-shim.md
```diff
@@ -0,0 +1,157 @@
...
+
+## File layout
+
+```
+src/spock_bucompat_5x.c       -- the entire mechanism (~325 lines)
+src/spock.c                   -- one extern decl + one call from _PG_init
+sql/spock--6.0.0-devel.sql    -- (no shim-related DDL; security-label
+                                  provider is registered in C)
+docs/internals-doc/binary-upgrade-compat-shim.md  -- this file
+```
```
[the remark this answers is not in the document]
danolivo: I removed this file

patches/16/pg16-000-spock-patchset-version.diff
```diff
@@ -0,0 +1,32 @@
...
++++ b/src/include/miscadmin.h
+@@ -510,4 +510,11 @@
+ /* in executor/nodeHash.c */
+ extern size_t get_hash_memory_limit(void);
+ 
++/*
++ * Spock core-patchset identity.  Bump the version when the patchset
++ * changes in a way visible to the extension binary.
++ */
++#define SPOCK_CORE_PATCHSET_VERSION 1
```
mason-sharp, line 17: I think it is smart idea in general. It does replace one patch with another though (but smaller). I am thinking one day we could have a spock-lite with reduced functionality that would work with plain Postgres. I will think about this some more.

sql/spock--5.0.0.sql
```diff
@@ -0,0 +1,790 @@
+\echo Use "CREATE EXTENSION spock" to load this file. \quit
```
mason-sharp, line 1: I don't think we need this file.

If users first start with 6.0.0, they would use spock--6.0.0.sql. 

If users have 5.0.x, then they are at some version >= 5.0.0 already.

sql/spock--6.0.0-devel.sql
```diff
@@ -18,7 +18,8 @@ CREATE TABLE spock.node_interface (
 
 CREATE TABLE spock.local_node (
     node_id oid PRIMARY KEY REFERENCES node(node_id),
-    node_local_interface oid NOT NULL REFERENCES node_interface(if_id)
+    node_local_interface oid NOT NULL REFERENCES node_interface(if_id),
+    node_version int4 NOT NULL DEFAULT 0
```
mason-sharp: A version check is a good idea, but can't we use `pg_extension.extversion` and simplify this? It could still be fetched in `get_local_node()` and catch when the .so does not match the last recorded version from `CREATE EXTENSION` or `ALTER EXTENSION`. Perhaps `spock_manager_main()` could check, and delay to avoid too many error messages in the log.

src/spock_bucompat_5x.c
```diff
@@ -0,0 +1,449 @@
+/*-------------------------------------------------------------------------
+ *
+ * spock_bucompat_5x.c
+ *		Binary-upgrade compatibility: spock 5.x -> 6.x.
```
mason-sharp: I like this, but I wonder if it would be simpler to just require on an upgrade (if we loosen the patch set versioning) that users must run a SQL upgrade script and then do a regular `pg_upgrade` after.

OTOH, if we defer until PG 19 only, this file as part of the upgrade process seems very useful/required.

Just adding comments while thinking about it.

mason-sharp (commented):
I did an initial pass and added some comments.

patches/16/pg16-000-spock-patchset-version.diff
danolivo: @mason-sharp ,
It is trivial to guard our patches and the corresponding Spock machinery. That enables Spock to be used with upstream Postgres. But for production use, at least one patch (logical clock) is a requirement. So, for the safety and predictability, we should have a mechanism to detect patched postgres.

sql/spock--5.0.0.sql
danolivo: Done. Although it is out of community coding standards, it is legal. So, I removed the base SQL script.

sql/spock--6.0.0-devel.sql
danolivo: The pg_extension.extversion doesn't simplify. Our purpose is to determine which Spock version was installed when the node was created. That's crucial because different versions can leave distinct fingerprints in the system. If you know how to avoid it and guarantee the upgrade, please share the details.

src/spock_bucompat_5x.c
danolivo: Right now, we keep things simple and don't require users to do anything beyond the standard pg_upgrade command. Not sure that the additional step is safer. One more reason: a separate module might be easily cut from the codebase when we decide not to support upgrades from 5.x anymore.

sql/spock--6.0.0-devel.sql
mason-sharp: If we run `ALTER EXTENSION UPDATE` it will update both `node_version` here as well as `pg_extension.extversion`. Before running it, these two value would be the same I think.

mason-sharp:
Will take a different approach

PR #429, opened 2026-04-23
Update sub_show_status() reference with status values

susan-pgedge:
This resolves spoc-361

Add a STATUS VALUES table documenting all five possible subscription states (replicating, initializing, disabled, down, unknown) derived from the C source. Fix heading markup, wrap synopsis in a fenced SQL code block, and add introductory sentences before examples.


mason-sharp:
Looks like there are merge conflicts.

mason-sharp (approved):

PR #434, opened 2026-04-27
Update release notes for 5.0.7

mason-sharp:


susan-pgedge (approved):
Looks good to me!  Do we want to open a ticket for each "Operational Improvement" moving forward so we are sure we catch each behavior change in the documentation?

docs/spock_release_notes.md
```diff
@@ -16,30 +18,50 @@
...
+  `remove_node` to fail when looking up subscriptions by name (SPOC-503). A
+  new `spock.gen_sub_name(provider_node, subscriber_node)` helper is now used
+  in place of inline name concatenations.
+* ZODAN: the `remove_node` cleanup loop tried to drop subscriptions by
+  guessing names and nodes, which broke whenever `cross_wire` and `add_node`
+  used different naming conventions. It now performs a per-node `DROP` of
+  every subscription read directly from the `spock.subscription` catalog.
 
+### Operational Improvements
+* ZODAN now uses the 5-argument form of `wait_for_sync_event` with explicit
```
rasifr: wait_for_sync_event has 4 args

mason-sharp:
> Looks good to me! Do we want to open a ticket for each "Operational Improvement" moving forward so we are sure we catch each behavior change in the documentation?

Yes, if applicable.

docs/spock_release_notes.md
mason-sharp: Good catch, that one slipped in there.

PR #436, opened 2026-04-28
zodan: fix Phase 3 catchup wait

danolivo:
Emit `spock.sync_event()` on rec **after** slot creation. By WAL ordering, its commit LSN is provably greater than the slot's `consistent_point`, which gives `remote_commit_lsn` a reachable target. Reaching it proves, by apply commit-order monotonicity, that every rec commit with `commit_lsn ≤ slot_consistent_point` is applied on src and visible to the COPY snapshot.

The single post-slot sync_event also replaces the prior pre-slot one — Phase 8's `wait_for_sync_event` on `new` consumes the same stored LSN, and a later target is strictly safer than an earlier one. One sync_event per rec instead of two.

The 3-minute wait timeout is promoted from a swallowed `WARNING + EXIT` to `RAISE EXCEPTION`. With a deterministic target in the WAL, a hang is a real failure that must abort `add_node` rather than risk data loss.

### `spock.sync_event(transactional boolean DEFAULT false)`

The fix relies on the apply-progress behaviour introduced in fdd9587: non-transactional message + WAL flush, with `replorigin_session_advance()` called explicitly on the apply side so `pg_replication_origin_status` reflects the marker's LSN immediately rather than only when an unrelated commit comes through.

To preserve that as the default while still letting callers who need ordering with surrounding DML opt in, `spock.sync_event()` now takes an optional `transactional` boolean argument:

- `false` (default) — non-transactional message + flush (Mason's behaviour from fdd9587). The marker is durable regardless of caller transaction outcome; waiters cannot hang because of a caller-side rollback. Required for any caller that polls progress against the returned LSN.
- `true` — buffered with the surrounding transaction, emitted at commit time. `wait_for_sync_event` then implies that all DML committed before the marker has also been applied. If the caller rolls back, the marker is discarded and any waiter hangs until its own timeout fires.

The trade-off is documented in the function comment.

### Naming cleanup

The local previously called `_commit_lsn` (and the matching `temp_sync_lsns.commit_lsn` column) is renamed to `_slot_lsn` throughout `samples/Z0DAN/zodan.sql`. The value held the LSN returned by `pg_create_logical_replication_slot`, which is the slot's `consistent_point` and not a commit LSN. Same rename for the diagnostic local in `check_commit_timestamp_and_advance_slot`.

### Migration impact

`spock.sync_event()` changed signature from `()` to `(transactional boolean DEFAULT false)`. Existing callers continue to work via the default. The `5.0.6 → 6.0.0-devel` upgrade path drops the zero-arg signature before recreating it to avoid two overloads competing for zero-arg resolution after `ALTER EXTENSION UPDATE`.

The `5.0.8 → 6.0.0-devel` upgrade path needs the same treatment — that migration script is still WIP locally and not part of this PR. To be added in a follow-up before 6.0.0 ships.

### Files changed

- `samples/Z0DAN/zodan.sql` — Phase 3 restructure (slot before sync_event, single post-slot sync_event), wait against `remote_commit_lsn`, `RAISE EXCEPTION` on timeout, `_slot_lsn` rename, comment updates.
- `src/spock_functions.c` — `transactional` arg threaded into `spock_create_sync_event`; DEBUG1 message no longer hard-codes "non-transactional"; function comment rewritten.
- `sql/spock--6.0.0-devel.sql` — new signature in canonical install.
- `sql/spock--5.0.6--6.0.0-devel.sql` — DROP old + CREATE new on upgrade.

### DIFF
--- a/samples/Z0DAN/zodan.sql
+++ b/samples/Z0DAN/zodan.sql
@@ -1338,7 +1338,8 @@ DECLARE
     dbname             text;
     slot_name          text;
 	sub_name           text;
-    _commit_lsn        pg_lsn;
+    _slot_lsn          pg_lsn;
+    _catchup_lsn       pg_lsn;
 BEGIN
     RAISE NOTICE 'Phase 3: Creating disabled subscriptions and slots';
 
@@ -1349,7 +1350,7 @@ BEGIN
     CREATE TEMP TABLE IF NOT EXISTS temp_sync_lsns (
         origin_node text PRIMARY KEY,
         sync_lsn text NOT NULL,
-        commit_lsn pg_lsn
+        slot_lsn pg_lsn
     );
 
     -- Check if there are any "other" nodes (not source, not new)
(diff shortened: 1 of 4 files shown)


mason-sharp:
Adding a couple of notes:
- This PR decreases the time on the 009_zodan_add_remove_nodes and 012_zodan_basics tests. A slowdown was introduced with the following PR: https://github.com/pgEdge/spock/pull/392.
- This PR is passing the long-running 011 test in a manually run github action.

ibrarahmad (approved):
LGTM

PR #444, opened 2026-04-29
Fix apply_delay dangling pointer in subscription_fromtuple.

ibrarahmad:


mason-sharp (approved):

mason-sharp:
The Codacy issue was a false positive. It was not getting propagated here.

PR #446, opened 2026-04-29
zodan: fix Phase 3 catchup wait; parameterise spock.sync_event()

mason-sharp:
Phase 3 of zodan add_node must ensure the source node has applied every change from each "other" rec up to the new slot's consistent_point before the Phase 5 src->new COPY runs.  Otherwise rows in the interval (src_apply_progress, slot_consistent_point] are missing from the COPY and the disabled rec->new slot picks up from slot_consistent_point forward, so those rows silently never reach new.

The prior wait compared spock.progress.remote_commit_lsn (or, more recently, received_lsn) on src against the slot's consistent_point. Neither worked.  The slot LSN typically lands on a RUNNING_XACTS record well past the last actual commit on rec, so on a quiet cluster remote_commit_lsn never converges and the wait times out.  Switching to received_lsn made the wait converge fast but silently broke the invariant: received_lsn is bumped by keepalive messages reporting the remote walsender's flush position, so it can race past the slot LSN while zero rec changes have been applied on src.

Fix by emitting spock.sync_event() on rec *after* slot creation.  Its commit LSN is provably greater than the slot's consistent_point by WAL ordering, which gives remote_commit_lsn a reachable target.  Reaching it proves, by apply commit-order monotonicity, that every rec commit with commit_lsn <= slot_consistent_point is applied on src and visible to the COPY snapshot.  The same LSN replaces the pre-slot sync_event that Phase 8 wait_for_sync_event consumed on new -- a later sync target is strictly safer than an earlier one.

(cherry picked from commit b7fc7022482e7d2caa732fcc0505d342e8e7eaa6)

### DIFF
--- a/samples/Z0DAN/zodan.sql
+++ b/samples/Z0DAN/zodan.sql
@@ -1338,7 +1338,8 @@ DECLARE
     dbname             text;
     slot_name          text;
 	sub_name           text;
-    _commit_lsn        pg_lsn;
+    _slot_lsn          pg_lsn;
+    _catchup_lsn       pg_lsn;
 BEGIN
     RAISE NOTICE 'Phase 3: Creating disabled subscriptions and slots';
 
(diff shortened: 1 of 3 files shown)


danolivo (approved):
LGTM

PR #447, opened 2026-04-30
Auto-add pre-existing tables to default repset on node_create.

ibrarahmad:
When CREATE TABLE runs before spock.node_create(), the autoddl hook silently no-ops because no local node exists yet, leaving the table out of every replication set so cross-wire propagates the schema but never the data.  Retro-add user tables in node_create, gated on the existing spock.include_ddl_repset GUC.


mason-sharp:
Expected behavior, closing.

PR #448, opened 2026-04-30
SPOC-527: add 024_node_id_collision for duplicate node_id behaviour

danolivo:
## Summary
Add a TAP test that exercises Spock's behaviour when two independently-created nodes end up sharing a `node_id`. The id is currently generated locally as `hash_any(name) & 0xffff` at `node_create` time, with no cluster-wide coordination — so this collision is reachable in practice (cluster splits, geo-migrations, DR rehydration). The test documents what happens today, before any negotiation protocol lands.
## What's covered
The test creates a 2-node cluster (`node_create` only, no cross-wire), then tampers with n2's catalog to give it n1's id. Tampering navigates two FK constraints: `node_interface.if_nodeid` and `replication_set.set_nodeid` follow via `ON UPDATE CASCADE`; `local_node.node_id` has no cascade and is updated manually under `SET session_replication_role = replica` to bypass the FK trigger. The PRIMARY KEY on `spock.node(node_id)` is enforced by the unique index, so this works only because we tamper pre-attach when each catalog still holds a single row.
With the colliding state in place, `sub_create` is attempted from both directions. The test asserts:
- both attempts fail with `duplicate key value violates unique constraint "node_pkey"` (the existing PK catches the collision when `create_node(origin)` at `src/spock_functions.c:503` tries to insert the remote-rep row);
- both `spock.node` catalogs still hold exactly one row (the failed transactions rolled back atomically);
- `spock.subscription` is empty on both sides (no orphan subscription rows pointing at a colliding id);
- n1's seed table was not synced to n2 (sub_create was invoked with `sync_structure=true`, `sync_data=true` — the PK trip has to happen before the copy phase or n2 would be left with an orphaned populated table).
## Out of scope
The cluster-merge case where a colliding id sits on a third-party peer (one the joining node hasn't subscribed to directly) produces silent misattribution rather than a PK trip. That requires a 3-node forwarding scenario and is left for the upcoming negotiation-protocol work.

### DIFF
--- a/tests/tap/schedule
+++ b/tests/tap/schedule
@@ -43,6 +43,7 @@ test: 018_forward_origins
 test: 018_failover_slots
 test: 019_stale_fd_epoll_after_conn_death
 test: 022_rmgr_progress_post_checkpoint_crash
+test: 024_node_id_collision
 # Upgrade schema match test (builds from source, slow):
 #test: 018_upgrade_schema_match
 #
--- a/tests/tap/t/024_node_id_collision.pl
+++ b/tests/tap/t/024_node_id_collision.pl
@@ -1,0 +1,178 @@
+use strict;
+use warnings;
+use Test::More;
+use lib '.';
+use SpockTest qw(create_cluster destroy_cluster system_or_bail command_ok
+                 get_test_config scalar_query psql_or_bail);
+
+# =============================================================================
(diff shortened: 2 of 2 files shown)


mason-sharp (approved):
Just a test; ok for 6.0.

PR #451, opened 2026-05-04
Avoid using TopMemoryContext

mason-sharp:
* Eliminate TopMemoryContext usage for spilled and spill-read entries. Entries now live in ApplyReplayContext and are cleaned up by its reset. Spill-read entries during replay are freed immediately after replication_handler, bounding replay memory to one entry at a time.

* Fix int overflow on apply_replay_bytes.  Was int, wrapped negative at ~2GB causing the spill threshold check to silently fail to activate spilling for very large in-memory queues.  Changed to int64.

* Simplify apply_replay_queue_append_entry signature from double pointers to direct pass-through, since the entry no longer needs to be relocated across memory contexts when spilled.

* Track the in-flight spilled PQ buffer via apply_replay_pending_pqfree so it can be freed if replication_handler throws ERROR before the caller's PQfreemem runs.  Defensive guard against malloc leaks on the exception path.

* Cleaner separation of stream-path and replay-path entry cleanup in the main apply loop.

* Fix mis-indented ConfigReloadPending check.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -135,13 +135,15 @@ static ApplyReplayEntry * apply_replay_tail = NULL;
 static ApplyReplayEntry * apply_replay_next = NULL;
 
 /* Total bytes of libpq-allocated data held by in-memory queue entries */
-static int	apply_replay_bytes = 0;
+static int64 apply_replay_bytes = 0;
 
 static bool apply_replay_mode = false;		/* true when replaying */
 static BufFile *apply_replay_spill_file = NULL;
 static bool apply_replay_spilling = false;
 static int	apply_replay_spill_count = 0;
 static int	apply_replay_spill_read = 0;
+/* In-flight spilled PQ buffer; freed by reset if ERROR skips PQfreemem. */
+static char *apply_replay_pending_pqfree = NULL;
 
 /*
  * A message counter for the xact, for debugging. We don't send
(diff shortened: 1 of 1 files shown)


src/spock_apply.c
```diff
@@ -135,13 +135,15 @@ static ApplyReplayEntry * apply_replay_tail = NULL;
...
-static int	apply_replay_bytes = 0;
+static int64 apply_replay_bytes = 0;
 
 static bool apply_replay_mode = false;		/* true when replaying */
 static BufFile *apply_replay_spill_file = NULL;
 static bool apply_replay_spilling = false;
 static int	apply_replay_spill_count = 0;
 static int	apply_replay_spill_read = 0;
+/* In-flight spilled PQ buffer; freed by reset if ERROR skips PQfreemem. */
+static char *apply_replay_pending_pqfree = NULL;
```
danolivo, line 146: There is a reason we declined such an approach in Spock before: having free/reset machinery at two levels of the code hierarchy is hard to support and tends to be buggy. For example, this specific implementation doesn't free the `entry` structure itself after the apply.
One more issue - if spilling fails for some reason, it causes a memory leak of the pq-allocated data.

Another reason why we used TopMemoryContext is to highlight the general memory management issues: MessageContext is a core Context and can't be reallocated/reset by an extension. Multiple context switches are also not nice.

The better approach is to use a SpillMemoryContext that lives at the same level as the message-receiving code, more strictly caps memory usage, and does not depend on the COMMIT command. But it should be introduced after (or with) the memory management refactoring procedure.

danolivo (changes requested):

mason-sharp:
Closing. While I think we should avoid TopMemoryContext, a larger refactor also makes sense.

PR #464, opened 2026-05-08
Arbitrary improvements (take 2)

danolivo:
## Summary

A grab-bag of independent cleanups in `contrib/spock`. None of them change user-visible behaviour; each commit stands on its own and can be reviewed separately.

- **PG18 compatibility for `--exclude-extension`.** The flag is new in Postgres 18, so its use is now gated behind a version check rather than relying on always-present API.
- **Drop Win32 support.** Removes the Windows-only command exec, temp-path resolution, argument quoting, and associated typedefs. Spock has been POSIX-only in practice — the dead paths just complicated the build matrix and the typedef list.
- **`spock_disable_subscription` warning comment.** Documents that the function's effect can be silently lost if the caller forgets to commit, which has caused coding errors before.
- **`spock--6.0.0-devel.sql` UI tidy-up.** Regroups the head install script into feature sections and normalises every `CREATE FUNCTION` to a single multi-line shape, so the script reads as a user-facing API rather than insertion-order history. Upgrade scripts are untouched.


src/spock_sync.c
```diff
@@ -166,7 +166,7 @@ get_pg_executable(char *cmdname, char *cmdbuf)
  * The sub->skip_schema typically doesn't include our globally-skipped objects.
  * So, don't care about duplicates.
  */
-
+#if PG_VERSION_NUM >= 180000
```
mason-sharp: I hadn't noticed --exclude-extension before, but it looks like it got added in Postgres 17.

danolivo: @mason-sharp , done.

mason-sharp (approved):

PR #465, opened 2026-05-08
Fix missing schema name in spock_version and spock_version_num docs

zaidshabbir25:
Fix missing schema name in spock_version and spock_version_num docs


susan-pgedge (approved):
Looks good to me!

PR #469, opened 2026-05-12
    Remove the PostgreSQL native failover slot work from 5.x.

ibrarahmad:


mason-sharp (approved):
Will do a follow up to update the release notes

PR #475, opened 2026-05-20
Make heap apply retry loop configurable via spock.read_retry_count

ibrarahmad:
spock_apply_heap_update() and spock_apply_heap_delete() hardcoded the retry count at 5 when the local tuple could not be found.  Expose this as a GUC (default 5, range 0..100, PGC_SIGHUP) so deployments hitting out-of-order arrivals can tune the loop without rebuilding.

### DIFF
--- a/src/spock.c
+++ b/src/spock.c
@@ -151,6 +151,7 @@ int			restart_delay_default;
 int			restart_delay_on_exception;
 int			spock_replay_queue_size;
 int			spock_pause_timeout = 10;	/* seconds to wait for apply workers to pause */
+int			spock_read_retry_count = 5;	/* heap update/delete: retries when local tuple is missing */
 bool		check_all_uc_indexes = false;
 bool		spock_enable_quiet_mode = false;
 int			log_origin_change = SPOCK_ORIGIN_NONE;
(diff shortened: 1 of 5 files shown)


src/spock.c
```diff
@@ -1195,6 +1196,26 @@ _PG_init(void)
...
 
+	DefineCustomIntVariable("spock.read_retry_count",
+							"Number of times the apply worker re-reads the local "
+							"relation when a row targeted by a remote UPDATE or "
+							"DELETE is not yet visible",
+							"On each retry the apply worker waits for any "
+							"concurrently-applying transaction to finish, then "
+							"searches the local relation again. Set to 0 to disable "
+							"retries (the row-missing path runs immediately). "
+							"Used in spock_apply_heap_update and "
```
mason-sharp: This is a little bit too verbose, no need to mention which functions.  Can you please shorten? Something like the first 4 lines and the "Set to 0" sentence?

mason-sharp (changes requested):

danolivo (approved):
LGTM

mason-sharp (changes requested):
Can you also update the docs for the new GUC?

mason-sharp (approved):

PR #477, opened 2026-05-20
Add spock.log_verbosity and spock.apply_change_logging GUCs

ibrarahmad:
log_verbosity (normal|verbose) promotes spock DEBUG1/DEBUG2 to LOG. apply_change_logging (none|key_only|verbose) emits per-change JSON from the apply worker: action/schema/table/pk/origin/commit_ts in key_only, plus old/new row data in verbose.  DDL is logged in both non-none modes.  Both GUCs are PGC_SUSET (SET / SIGHUP reload).

TAP coverage in tests/tap/t/044_apply_change_logging.pl - 24/24.

SPOC-551

### DIFF
--- a/src/spock.c
+++ b/src/spock.c
@@ -141,7 +141,8 @@ static const struct config_enum_entry readonly_options[] = {
 
 static const struct config_enum_entry log_verbosity_options[] = {
 	{"normal", SPOCK_LOG_VERBOSITY_NORMAL, false},
-	{"verbose", SPOCK_LOG_VERBOSITY_VERBOSE, false},
+	{"debug1", SPOCK_LOG_VERBOSITY_DEBUG1, false},
+	{"debug2", SPOCK_LOG_VERBOSITY_DEBUG2, false},
 	{NULL, 0, false}
 };
 
@@ -1083,9 +1084,10 @@ _PG_init(void)
 
 	DefineCustomEnumVariable("spock.log_verbosity",
 							 gettext_noop("Sets the verbosity of spock-specific log output."),
-							 gettext_noop("When 'verbose', spock's own DEBUG1 and DEBUG2 messages are "
-										  "promoted to LOG level so they appear in standard server logs "
-										  "without changing log_min_messages globally."),
+							 gettext_noop("Promotes spock's own debug messages to LOG so they appear in "
+										  "standard server logs without changing log_min_messages globally. "
+										  "'normal' (default) promotes nothing; 'debug1' promotes spock's "
+										  "DEBUG1 messages; 'debug2' promotes both DEBUG1 and DEBUG2."),
 							 &spock_log_verbosity,
 							 SPOCK_LOG_VERBOSITY_NORMAL,
 							 log_verbosity_options,
--- a/src/spock.c
+++ b/src/spock.c
@@ -139,6 +139,19 @@ static const struct config_enum_entry readonly_options[] = {
 	{NULL, 0, false}
 };
 
+static const struct config_enum_entry log_verbosity_options[] = {
+	{"normal", SPOCK_LOG_VERBOSITY_NORMAL, false},
+	{"verbose", SPOCK_LOG_VERBOSITY_VERBOSE, false},
+	{NULL, 0, false}
+};
+
+static const struct config_enum_entry apply_change_logging_options[] = {
+	{"none", SPOCK_APPLY_CHANGE_LOG_NONE, false},
+	{"key_only", SPOCK_APPLY_CHANGE_LOG_KEY_ONLY, false},
+	{"verbose", SPOCK_APPLY_CHANGE_LOG_VERBOSE, false},
+	{NULL, 0, false}
+};
+
 bool		spock_synchronous_commit = false;
 char	   *spock_temp_directory = "";
 static char *spock_temp_directory_config;
@@ -157,6 +170,8 @@ bool		check_all_uc_indexes = false;
 bool		spock_enable_quiet_mode = false;
 int			log_origin_change = SPOCK_ORIGIN_NONE;
 int			spock_apply_idle_timeout = 300;
+int			spock_log_verbosity = SPOCK_LOG_VERBOSITY_NORMAL;
+int			spock_apply_change_logging = SPOCK_APPLY_CHANGE_LOG_NONE;
 
 static emit_log_hook_type prev_emit_log_hook = NULL;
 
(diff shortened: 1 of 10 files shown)


src/spock.c
```diff
@@ -1058,6 +1100,30 @@ _PG_init(void)
...
+										  "promoted to LOG level so they appear in standard server logs "
+										  "without changing log_min_messages globally."),
+							 &spock_log_verbosity,
+							 SPOCK_LOG_VERBOSITY_NORMAL,
+							 log_verbosity_options,
+							 PGC_SUSET, 0,
+							 NULL, NULL, NULL);
+
+	DefineCustomEnumVariable("spock.apply_change_logging",
+							 gettext_noop("Controls JSON logging of changes applied by the apply worker."),
```
rasifr, line 1098: I don't think we need both `spock.log_verbosity` and `spock.apply_change_logging`. A single spock.debug_log_level using the stock server_message_level_options enum should be sufficient: increasing the DEBUG level increases verbosity exactly like Postgres itself.

That also removes the need for the new SPOCK_DEBUG1/SPOCK_DEBUG2 macros — every existing ereport(DEBUG1/2, …) is already gated by log_min_messages, so the GUC simply controls the level at which spock's own debug output is emitted.

src/spock_apply.c
```diff
@@ -1576,6 +1581,11 @@ handle_update(StringInfo s)
 		spock_apply_heap_update(rel, hasoldtup ? &oldtup : &newtup, &newtup);
 	}
 
+	if (!failed)
+		spock_log_apply_change("UPDATE", rel,
+							   hasoldtup ? &oldtup : NULL, &newtup,
```
rasifr, line 1651: Should the change-log record really be gated by !failed?  When user turns this on they almost certainly want to see the conflict and exception cases — those are the interesting ones. As written, it silently drops exactly the records you'd enable it to capture. I'd either always emit (with "incoming"|"applied"|"conflict"|"skipped" field, can be a generic string) or hook into spock_report_conflict()/log_insert_exception. 

Same comment applies to the DDL path in handle_sql — it logs before execution, so a failed DDL still produces an "action":"DDL" line that looks applied.

src/spock_change_log.c
```diff
@@ -0,0 +1,290 @@
...
+#include "spock_relcache.h"
+#include "spock_change_log.h"
+
+/*
+ * Append a JSON-quoted string.  Handles \\, \", and control chars below 0x20.
+ */
+static void
+append_json_string(StringInfo buf, const char *str)
+{
+	appendStringInfoChar(buf, '"');
```
rasifr: src/spock_jsonb_utils.c already has most of the tuple->JSON functionality needed here (and already makes use of escape_json()); it should be enhanced where it falls short rather than duplicated.

mason-sharp: +1

src/spock_apply.c
mason-sharp: Should we instead simply always log for IUD before attempting the spock_apply_heap_*? Then we will always have the attempted change before it fails or succeeds.

The DDL path logs before it attempts to change anything already.

src/spock.c
mason-sharp: Fair point, but I actually like this to have Spock-specific logging to cut down on noise and use less disk space.

mason-sharp (changes requested):

rasifr (commented):
- Although I still favor single GUC, even accepting two separate GUCs, the new SPOCK_DEBUG1/SPOCK_DEBUG2 macros are unnecessary complexity. Every existing ereport(DEBUG1/2, …) is already gated by log_min_messages — that is exactly the mechanism Postgres provides for controlling what debug output reaches the log. Wrapping that with a spock-specific level macro that duplicates the concept adds indirection without adding capability. 
If the goal is to promote spock debug messages to LOG without touching log_min_messages globally, the GUC alone can do that at the call site with a simple helper function.

- spock_jsonb_utils.c already provides spock_tuple_to_json_cstring() — the same conceptual operation this PR implements again in append_row_json. The right approach is to extend spock_jsonb_utils.c to cover the additional requirements (such as PK-only filtering etc) rather than introduce a parallel set of JSON helpers in a new file.

- The removal of the entire Win32 exec_cmd path (~324 lines) and the PG_VERSION_NUM guard change from >= 180000 to >= 170000 for build_exclude_extension_string have no connection to the GUC feature being added here. These should be in a separate PR

mason-sharp:
> * Although I still favor single GUC, even accepting two separate GUCs, the new SPOCK_DEBUG1/SPOCK_DEBUG2 macros are unnecessary complexity. Every existing ereport(DEBUG1/2, …) is already gated by log_min_messages — that is exactly the mechanism Postgres provides for controlling what debug output reaches the log. Wrapping that with a spock-specific level macro that duplicates the concept adds indirection without adding capability.
>   If the goal is to promote spock debug messages to LOG without touching log_min_messages globally, the GUC alone can do that at the call site with a simple helper function.

OK, now I understand what you meant. `spock.apply_change_logging` could promote to LOG if the mode is not NONE. @ibrarahmad what do you think?

> * spock_jsonb_utils.c already provides spock_tuple_to_json_cstring() — the same conceptual operation this PR implements again in append_row_json. The right approach is to extend spock_jsonb_utils.c to cover the additional requirements (such as PK-only filtering etc) rather than introduce a parallel set of JSON helpers in a new file.

@ibrarahmad I agree with Asif.

> * The removal of the entire Win32 exec_cmd path (~324 lines) and the PG_VERSION_NUM guard change from >= 180000 to >= 170000 for build_exclude_extension_string have no connection to the GUC feature being added here. These should be in a separate PR

I think he just needs to rebase, that was already removed.

src/spock_apply.c
```diff
@@ -1410,6 +1411,10 @@ handle_insert(StringInfo s)
 		spock_apply_heap_insert(rel, &newtup);
 	}
 
+	if (!failed)
+		spock_log_apply_change("INSERT", rel, NULL, &newtup,
+							   remote_origin_name);
```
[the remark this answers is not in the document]
ibrarahmad: Fixed in 66ba3cd. The log call now runs before the apply attempt and the !failed guard is gone, so skipped IUDs under TRANSDISCARD and SUB_DISABLE still get logged.

src/spock.c
ibrarahmad: Going with the two-GUC setup so change logging can be turned on without raising spock's debug verbosity for everything else, matching the preference in the reply below.

src/spock_apply.c
ibrarahmad: Moved spock_log_apply_change ahead of apply_api.do_* in 66ba3cd and dropped the !failed gate, so every IUD that reaches the apply path is logged whether it succeeds, fails, or is skipped. DDL keeps the existing pre-execute log.

src/spock_change_log.c
ibrarahmad: 90e60e5 moves the Datum-to-JSON-string helper out of spock_change_log into spock_jsonb_utils as spock_append_datum_as_json_string, so the change-log emitter no longer carries its own copy.

src/spock_apply.c
ibrarahmad: Done that way in 66ba3cd, so every IUD logs before spock_apply_heap_*.

src/spock.c
ibrarahmad: Agreed, keeping the Spock-specific GUC.

src/spock_change_log.c
ibrarahmad: Helper now lives in spock_jsonb_utils as of 90e60e5.

mason-sharp:
Gave this a rebase because of the extra windows code removal changes

rasifr (approved):

mason-sharp (approved):

PR #480, opened 2026-05-24
spock: revalidate dboid before registering a manager worker

ibrarahmad:
Between the pg_database seq-scan and spock_worker_register() another backend can DROP DATABASE or set datconnlimit = -2, leaving the manager worker to FATAL during attach and sustaining a respawn loop under heavy tenant-DB churn.

SPOC-567

### DIFF
--- a/tests/tap/t/103_manager_worker_dboid_race.pl
+++ b/tests/tap/t/103_manager_worker_dboid_race.pl
@@ -1,0 +1,138 @@
+use strict;
+use warnings;
+use Test::More;
+use POSIX ();
+use lib '.';
+use lib 't';
+use SpockTest qw(
+    create_cluster destroy_cluster
+    get_test_config system_or_bail
+);
+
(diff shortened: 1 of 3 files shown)


tests/tap/t/103_manager_worker_dboid_race.pl
```diff
@@ -0,0 +1,138 @@
...
+use lib 't';
+use SpockTest qw(
+    create_cluster destroy_cluster
+    get_test_config system_or_bail
+);
+
+# =============================================================================
+# Test 103: manager-worker dboid race during tenant-DB churn
+#
+# Reproduces the SUP-137 scenario:
```
mason-sharp: Let's avoid referring to internal ticket numbers in code, here and below.

mason-sharp (changes requested):

mason-sharp (approved):

PR #490, opened 2026-06-03
autoddl: keep ALTER TABLE from yanking tables out of custom repsets

rasifr:
With autoDDL and spock.include_ddl_repset both enabled, any ALTER TABLE
    on a table that lived in a user-defined replication set was being
    forcibly re-routed back into 'default' (or 'default_insert_only'),
    silently undoing the operator's repset placement.

    Apply ALTER stickiness rules in apply_repset_policy_for_reloid():

      - If the ALTER did not add or drop a PK / replica identity, leave
        repset membership entirely alone.
      - If the ALTER added a PK on a table in a custom repset, emit a
        NOTICE and leave membership alone. The operator can move it
        manually if they want full UPDATE/DELETE replication.
      - If the ALTER dropped the PK on a table in a custom repset that
        replicates UPDATE/DELETE, emit a WARNING and move it to
        default_insert_only. Without a replica identity, that custom
        repset would otherwise produce broken downstream replication.
      - Tables in the default repsets continue to be auto-re-routed as
        before.

### DIFF
--- a/src/spock_autoddl.c
+++ b/src/spock_autoddl.c
@@ -527,6 +527,31 @@ classify_alter_pkri_change(AlterTableStmt *atstmt, Relation targetrel)
 						has_add_pkri_cmd = true;
 				}
 				break;
+			case AT_AddColumn:
+				/*
+				 * "ALTER TABLE ... ADD COLUMN x int PRIMARY KEY" arrives as
+				 * AT_AddColumn with the PRIMARY KEY embedded in the
+				 * ColumnDef's inline constraints list, not as a separate
+				 * AT_AddConstraint command.
+				 */
+				if (cmd->def && IsA(cmd->def, ColumnDef))
+				{
+					ColumnDef  *col = (ColumnDef *) cmd->def;
+					ListCell   *cc;
+
+					foreach(cc, col->constraints)
+					{
+						Constraint *con = (Constraint *) lfirst(cc);
+
+						if (IsA(con, Constraint) &&
+							con->contype == CONSTR_PRIMARY)
+						{
+							has_add_pkri_cmd = true;
+							break;
+						}
+					}
+				}
+				break;
 			case AT_ReplicaIdentity:
 				/*
 				 * Could be adding or removing RI; let the post-state
--- a/src/spock_autoddl.c
+++ b/src/spock_autoddl.c
@@ -35,13 +35,35 @@
 #include "spock_output_plugin.h"
 #include "spock.h"
 
+/*
+ * Classification of a post-execution ALTER TABLE with respect to the
+ * table's primary key / replica identity. apply_repset_policy_for_reloid()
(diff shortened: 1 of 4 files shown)


src/spock_autoddl.c
```diff
@@ -113,190 +138,478 @@ spock_autoddl_process(PlannedStmt *pstmt,
...
+						if (IsA(con, Constraint) &&
+							con->contype == CONSTR_PRIMARY)
+						{
+							has_add_pkri_cmd = true;
+							break;
+						}
+					}
+				}
+				break;
+			case AT_ReplicaIdentity:
```
ibrarahmad, line 552: AT_ReplicaIdentity sets both add and drop flags here, so any ALTER ... REPLICA IDENTITY ... gets reclassified by post-state alone. A no-PK table moving to REPLICA IDENTITY FULL still has rd_replidindex invalid, so post_has_pk_or_ri is false and this returns PKRI_DROPPED, firing a spurious lost-its-primary-key WARNING even though FULL is a valid setup for UPDATE/DELETE.

src/spock_autoddl.c
```diff
@@ -113,190 +138,478 @@ spock_autoddl_process(PlannedStmt *pstmt,
...
+				/*
+				 * Could be adding or removing RI; let the post-state
+				 * disambiguate.
+				 */
+				has_add_pkri_cmd = true;
+				has_drop_pkri_candidate = true;
+				break;
+			case AT_DropConstraint:
+			case AT_DropColumn:
+				has_drop_pkri_candidate = true;
```
ibrarahmad, line 585: AT_DropColumn and AT_DropConstraint set has_drop_pkri_candidate without checking the name. A no-PK table that an admin placed in a custom UPDATE/DELETE repset, then dropping any unrelated CHECK or column, returns PKRI_DROPPED here, fires the lost-its-primary-key WARNING, and gets demoted to default_insert_only.

src/spock_autoddl.c
```diff
@@ -113,190 +138,478 @@ spock_autoddl_process(PlannedStmt *pstmt,
...
+ *		- Otherwise: leave membership alone.
+ *	- Else (CREATE, CREATE TABLE AS, ATTACH PARTITION, ALTER with no custom
+ *	  repset, or fall-through from the PK-dropped safety net):
+ *		- PK or replica identity present -> 'default' repset.
+ *		- No PK/RI -> 'default_insert_only'.
+ */
+static void
+apply_repset_policy_for_reloid(SpockLocalNode *node, Oid reloid,
+							   AlterTableStmt *atstmt,
+							   bool check_alter_stickiness)
```
ibrarahmad: check_alter_stickiness is set true exactly when atstmt is non-NULL, so the two are redundant. Drop the bool parameter and gate on atstmt != NULL inside the function.

ibrarahmad (approved):

mason-sharp:
Gave it a rebase

PR #491, opened 2026-06-03
spock_apply: do not send local LSN as publisher feedback on sync-standby

ibrarahmad:
The synchronous-standby hold queue stored the subscriber's local XactLastCommitEnd in RemoteSyncPosition.recvpos and later sent it to the publisher as the slot's confirmed flush position. The two LSNs belong to different WAL streams, so the publisher wrote a meaningless value into confirmed_flush_lsn -- typically far ahead of its own pg_current_wal_lsn. After a CNPG failover the promoted standby resumed from that bogus LSN, the publisher waited until its own WAL caught up, then started streaming from there. Every commit between the real origin position and the bogus LSN was silently dropped.

Split RemoteSyncPosition into local_commit_lsn (gating) and remote_recvpos / remote_writepos / remote_flushpos (feedback). append_feedback_position now takes both LSNs explicitly. get_feedback_position gates the release on local_commit_lsn <= WalSndCtl->lsn[SYNC_REP_WAIT_FLUSH] and returns the remote positions, so the publisher sees values from its own WAL stream.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -233,15 +233,37 @@ static ActionErrCallbackArg errcallback_arg;
 TransactionId remote_xid;
 
 /*
- * Structure of RemoteSyncPosition to Save the LSN in case of
- * Synchronous replica is attached
+ * Per-transaction feedback hold for the synchronous-replica path.
+ *
+ * When a local synchronous standby is configured (synchronous_standby_names
+ * is set), we cannot report a remote commit LSN back to the publisher until
+ * the corresponding local commit has been flushed on that sync standby --
+ * otherwise the publisher could advance its slot past changes that a
+ * fail-over of THIS node has not yet preserved.
+ *
+ * The two LSNs are not interchangeable:
+ *   - local_commit_lsn  is the subscriber's XactLastCommitEnd, used ONLY to
+ *                       compare against WalSndCtl->lsn[SYNC_REP_WAIT_FLUSH]
+ *                       to decide when the hold can be released.
+ *   - remote_recvpos / remote_writepos / remote_flushpos are positions in
+ *                       the publisher's WAL stream.  These are what we send
+ *                       in the standby-status feedback message; the
+ *                       publisher writes flushpos into the slot's
+ *                       confirmed_flush_lsn.
+ *
+ * A prior version of this struct stored only "recvpos" and reused it for
+ * both purposes -- the local LSN ended up being sent to the publisher,
+ * which then poisoned the slot's confirmed_flush_lsn with a value from a
+ * different WAL space.  After a failover, the new primary would resume from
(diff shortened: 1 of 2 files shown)


src/spock_apply.c
```diff
@@ -2768,55 +2791,81 @@ get_flush_position(XLogRecPtr *write, XLogRecPtr *flush)
...
+ *                        report (recv = write position in the publisher's
+ *                        WAL stream we have processed, flush/write are the
+ *                        equivalent positions returned by get_flush_position
+ *                        once the local commit is durable).
+ *
+ * Sending the local LSN as the remote position (the previous behaviour of
+ * this function) poisons the publisher's slot->confirmed_flush_lsn with a
+ * value from a different WAL space and silently drops every transaction
+ * after a fail-over until the publisher's own WAL grows past the bogus
+ * value.
```
mason-sharp: I think we can remove the last paragraph.

src/spock_apply.c
```diff
@@ -2768,55 +2791,81 @@ get_flush_position(XLogRecPtr *write, XLogRecPtr *flush)
...
- * by the synchronous replica.
+ * Walk the queue of sync-replica holds populated by append_feedback_position()
+ * and release any whose LOCAL commit has been flushed on the local sync
+ * standby.  For each released entry we overwrite the caller's *recvpos /
+ * *writepos / *flushpos with the entry's REMOTE positions -- those are
+ * what go into the standby-status reply to the publisher.  Mixing the two
+ * LSN spaces (the previous behaviour) caused the publisher to write a
+ * subscriber-local LSN into the slot's confirmed_flush_lsn, which after a
+ * fail-over made the new primary resume from a non-existent publisher
+ * position and silently drop every transaction in the window.
```
mason-sharp: I would remove the last sentence.

mason-sharp (commented):
Can you also add a test?

tests/tap/t/020_sync_standby_feedback_lsn.pl
```diff
@@ -0,0 +1,260 @@
...
+    SELECT bool_and(s.confirmed_flush_lsn <= pg_current_wal_lsn())
+    FROM pg_replication_slots s
+    WHERE s.slot_type = 'logical'
+      AND s.slot_name LIKE 'spk_${dbname}_n1_%'
+");
+$slot_ok =~ s/\s+//g;
+is($slot_ok, 't',
+    "publisher's confirmed_flush_lsn <= pg_current_wal_lsn (sync-standby feedback "
+  . "uses REMOTE LSN, not local)");
+
```
mason-sharp, line 228: The original bug was discovered after the standby was promoted and it did not receive some writes from the publisher. Should we add that?

mason-sharp (approved):

PR #493, opened 2026-06-08
spock_apply_heap: reuse ResultRelInfo indexes; free index predicate tree

ibrarahmad:
Iterate edata->targetRelInfo->ri_IndexRelationDescs instead of re-opening each index per row, and only set *indexoid on a match. Wrap RelationGetIndexPredicate in a scratch context so the tree is freed.


ibrarahmad (commented):
I think that look ok to me

mason-sharp (approved):

PR #497, opened 2026-06-10
spock_apply: surface root cause when a discard is not attributable to a row

rasifr:
TRANSDISCARD/SUB_DISABLE replay records the real error on the row that
    failed and points the other discarded rows at it via command_counter
    (cheaper than repeating the message on every row). When the failure is not
    attributable to a specific row, failed_action is 0 and that pointer dangles
    -- no row's counter is 0 -- so the captured cause was lost, e.g. "tuple
    discarded due to exception at command_counter 0".

    Extract the collateral-discard message into discard_collateral_message(),
    shared by the DML and SQL replay paths, and surface the captured
    initial_error_message when failed_action is 0 instead of an empty pointer.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -1211,6 +1211,41 @@ get_exception_behaviour_name(void)
 }
 
 /*
+ * Build the error_message for an entry discarded as collateral -- i.e. it was
+ * not itself the failing command (the caller passed a NULL message).  'what'
+ * names the discarded entry for the message: "tuple" on the DML path,
+ * "statement" on the SQL/DDL path (where no tuple is involved).
+ *
+ * Normally we just point the operator at the entry that carries the real error
+ * via its command_counter, which avoids repeating the same message on every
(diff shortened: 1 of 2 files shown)


danolivo (changes requested):
1. handle_commit increments this counter too. What if we fail somewhere in the middle of a commit, due to a deferred unique or FK constraint, as an example? Not sure if anything would go wrong (except incorrect reference in the exceptions table), but there should be a regression test to verify it is OK.
2. discard_collateral_message() dereferences exception_log_ptr[my_exception_log_index] with no guard; the SQL caller asserts the index, but the DML caller doesn't. An Assert(exception_log_ptr != NULL && my_exception_log_index >= 0) in the helper would match the paranoia level of the surrounding code.
3. "tuple discarded" - Does this make sense in the case of SQL path (DDL command)? What tuple has been discarded in such a case?

danolivo (approved):

PR #504, opened 2026-06-16
spock_conflict: fix spurious tiebreaker-equal WARNING in single-writer topology

rasifr:
conflict_resolve_by_timestamp() used get_node(local_origin_id) for loc_node,
where local_origin_id is the replication origin stamped on the existing local
tuple.  In a single-writer topology every row on the subscriber carries
origin = provider.id and incoming changes also have remote_origin = provider.id,
so both lookups resolved to the same SpockNode.  The same ambiguity arises from
a node_id hash collision: the id is a 16-bit hash of the node name, so two
distinct nodes can share the same id and again resolve to the same SpockNode.
In either case the tiebreaker comparison is trivially equal, firing a spurious
WARNING on every timestamp-tied conflict.

Use get_local_node() for loc_node so the tiebreaker always compares the
applying node against the remote sender, not the tuple's origin.  Add a
same-origin shortcut: when local_origin_id == remote_origin_id the change
shares the same source (e.g. the same row updated twice in one upstream
transaction), so there is no real conflict and remote is applied directly
without consulting the tiebreaker.

Detect node_id hash collisions at node creation time.

### DIFF
--- a/src/spock_node.c
+++ b/src/spock_node.c
@@ -178,6 +178,7 @@ create_node(SpockNode *node)
 	Datum		values[Natts_node];
 	bool		nulls[Natts_node];
 	NameData	node_name;
+	SpockNode  *existing;
 
 	if (get_node_by_name(node->name, true) != NULL)
 		elog(ERROR, "node %s already exists", node->name);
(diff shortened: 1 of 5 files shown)


src/spock_node.c
```diff
@@ -188,6 +189,19 @@ create_node(SpockNode *node)
...
+	 * hash of the node name, so two different names can produce the same
+	 * value.  A collision causes replication-origin ambiguity between the
+	 * two nodes.
+	 */
+	existing = get_node(node->id, true);
+	if (existing != NULL)
+		elog(ERROR,
+				"node id %u (computed from name \"%s\") collides with existing "
+				"node \"%s\"; rename this node to resolve the collision",
+				node->id, node->name, existing->name);
```
mason-sharp, line 203: 👍

mason-sharp (approved):

PR #510, opened 2026-06-18
node-level skip_schema — schemas a node never replicates

danolivo:
## Summary
Adds a node-level "skip schema" facility: a node lists schema names in its `spock.node.info->'skip_schema'` jsonb array, and those schemas are never replicated to or from that node. This is meant for node-local bookkeeping schemas (e.g. ACE's) that exist on a node but must stay out of replication, regardless of replication-set configuration.
The list is enforced at the three points where a schema's objects could otherwise enter replication.
## What changes
**Parse + repset gate** (`spock_node.c`, `spock_node.h`)
- `SpockNode.skip_schema` is populated from `info->'skip_schema'` when a node row is loaded.
- `EnsureRelationNotIgnored()` refuses to add a table in a skipped schema to any replication set. Because the output plugin already drops anything not in a replication set, this alone keeps new objects out.
**Structure sync** (`spock_sync.c`)
- `dump_structure()` passes `--exclude-schema` for each schema the local node (`sub->target`) has registered as skipped, so a fresh subscriber never recreates them from the provider's dump. The skip list rides in with the subscription, so no extra catalog lookup is needed.
**Publisher filter** (`spock_output_plugin.c`, `spock_output_plugin.h`)
- The output plugin drops row changes and TRUNCATEs whose relation lives in a skipped schema. This catches the transition case the repset gate cannot: a table already in a replication set when its schema is later skipped.
- The skip set is cached as resolved namespace OIDs, so the per-change test is a plain OID membership check — no syscache lookup or allocation in the hot path.
- The cache is reloaded lazily when it may have gone stale: a `spock.node` row change seen in the decoded stream (the list may differ), or a relcache invalidation (a namespace create/drop/rename/OID-reuse may change name→OID resolution). Reloads read only `spock.node` (a `user_catalog_table`, safe under the historic snapshot) rather than parsing the decoded tuple.
- The per-change special-case catalog checks (queue / replication_set / node) are gated behind a single spock-schema namespace comparison, so an ordinary user-table change rules them all out at once.

## Known limitation
`--exclude-schema` does not fully cover extensions: an extension living in a skipped schema is dumped as a global `CREATE EXTENSION ... WITH SCHEMA <skipped>`, which would recreate the schema on the target. Excluding it requires discovering the extension on the provider and emitting `--exclude-extension`. This is documented as an `XXX` in `dump_structure()` and intentionally not handled here.

### DIFF
--- a/include/spock_node.h
+++ b/include/spock_node.h
@@ -24,6 +24,8 @@ typedef struct SpockNode
 	Jsonb	   *info;
 	/* Fields contained in the info jsonb */
 	int32		tiebreaker;
+	List	   *skip_schema;	/* "skip_schema" array: schemas this node
+								 * never replicates (NIL when unset) */
 } SpockNode;
 
 typedef struct SpockInterface
--- a/include/spock_output_plugin.h
+++ b/include/spock_output_plugin.h
@@ -79,6 +79,9 @@ typedef struct SpockOutputData
 	/* List of SpockRepSet */
 	List	   *replication_sets;
 	RangeVar   *replicate_only_table;
+	/* Never-replicate schemas (info->'skip_schema') as namespace OIDs, reloaded
+	 * when the global skip_schema_valid flag is cleared. */
+	List	   *skip_schema;	/* list of namespace OIDs */
 } SpockOutputData;
 
 /*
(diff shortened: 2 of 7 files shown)


mason-sharp:
Used this PR instead: https://github.com/pgEdge/spock/pull/530

PR #511, opened 2026-06-19
surface real exception_log cause

mason-sharp:
spock_apply: surface real exception_log cause, not "unavailable"
Backport the exception_log message-quality work from main, adapted to
v5_STABLE.  The files have diverged enough that these do not cherry-pick
cleanly, so the logic was reapplied against the current v5_STABLE code:

  - ed17523 "exception_log: replace 'unavailable' with informative discard
    message" -- collateral rows (those not themselves the failing command)
    now carry a message naming the active behaviour and the command that
    failed, instead of the opaque placeholder "unavailable".
  - b32ef95 "spock_apply: surface root cause when a discard is not
    attributable to a row" -- when failed_action is 0 (e.g. a COMMIT-time
    failure) the captured root cause is surfaced instead of a dangling
    command_counter pointer.
  - 1f5738c "fix: clean up exception_log error-propagation code" -- folded
    in for the error-propagation fix the two above build on.

Adapting was non-trivial: v5_STABLE's log_insert_exception() dropped the
message via "(failed) ? errmsg : NULL", so the matching row's
initial_error_message never reached the log.  That is corrected here so the
real message actually lands.

On top of the backport:

  - Tag every captured/surfaced message with its SQLSTATE
    ([SQLSTATE xxxxx] ...), in exception_log and the "caught initial
    exception" log line.  exception_log has no sqlstate column, so carrying
    it inline is the cheapest way to make the root cause unambiguous (e.g.
    to tell a real constraint violation from a transient conflict).
  - Allocate the formatted messages in ApplyOperationContext (reset per
    row) rather than the long-lived TopTransactionContext current on the
    exception path, so logging every discarded row of a large transaction
    does not accumulate.

Diagnostic change only: discard/disable/LSN behaviour is unchanged.
Connection-class errors still rethrow before this path (https://github.com/pgEdge/spock/commit/b75cb4fc4646a31e3db63a624af1354fc58c13b2) and never
reach exception_log.

Test: 013_exception_handling.pl Part 5 updated to the new contract -- no
"unavailable"; failing row carries its SQLSTATE; bystander rows carry the
informative discard message.

### DIFF
--- a/src/spock_exception_handler.c
+++ b/src/spock_exception_handler.c
@@ -192,7 +192,6 @@ add_entry_to_exception_log(Oid remote_origin, TimestampTz remote_commit_ts,
 	 * defensive fallback so a future caller passing NULL degrades to a
 	 * placeholder rather than crashing.
 	 */
-	Assert(error_message != NULL);
 	values[Anum_exception_log_error_message - 1] =
 		CStringGetTextDatum(error_message != NULL ? error_message : "unavailable");
 	values[Anum_exception_log_retry_errored_at - 1] = TimestampTzGetDatum(GetCurrentTimestamp());
--- a/src/spock_exception_handler.c
+++ b/src/spock_exception_handler.c
@@ -186,10 +186,15 @@ add_entry_to_exception_log(Oid remote_origin, TimestampTz remote_commit_ts,
 		values[Anum_exception_log_ddl_user - 1] = CStringGetTextDatum(ddl_user);
 	}
 
-	if (error_message == NULL)
-		values[Anum_exception_log_error_message - 1] = CStringGetTextDatum("unavailable");
-	else
-		values[Anum_exception_log_error_message - 1] = CStringGetTextDatum(error_message);
+	/*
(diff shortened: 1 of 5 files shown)


src/spock_exception_handler.c
```diff
@@ -186,10 +186,15 @@ add_entry_to_exception_log(Oid remote_origin, TimestampTz remote_commit_ts,
...
-	else
-		values[Anum_exception_log_error_message - 1] = CStringGetTextDatum(error_message);
+	/*
+	 * All callers now build an informative message (the failing command's
+	 * error, or a collateral-discard message) before reaching here.  Keep a
+	 * defensive fallback so a future caller passing NULL degrades to a
+	 * placeholder rather than crashing.
+	 */
+	Assert(error_message != NULL);
+	values[Anum_exception_log_error_message - 1] =
```
rasifr: The Assert() here is not needed and should be removed.

rasifr (approved):

PR #521, opened 2026-07-09
[WIP] Physical to logical replica support

rasifr:


rasifr:
@coderabbitai review commit 051450c

rasifr:
@coderabbitai 

1. 
When an enabled subscription's apply worker starts, it calls replorigin_session_setup(originid) once, which sets replication_state->acquired_by = MyProcPid and holds it for the worker's entire lifetime (spock_apply.c:4647, no per-transaction release). Critically, replorigin_advance() — the same function maybe_advance_forwarded_origin() calls — checks that ownership first:

if (replication_state->acquired_by != 0)
    ereport(ERROR, (errcode(ERRCODE_OBJECT_IN_USE),
             errmsg("replication origin with ID %d is already active for PID %d", ...)));

So if sub_n3_n2's own worker is alive and owns that origin, a forwarding-path call to advance the same origin doesn't silently succeed or race on "latest LSN wins" — it throws ERRCODE_OBJECT_IN_USE immediately. 

Besides, whichever path apply's the LSN, and advances, at most it could duplicate/resend same or older transaction, there will be no data loss.

PR #523, opened 2026-07-09
Allow zodan add_node to work with lolor large object data.

mason-sharp:
Permit the new node to have lolor installed as long as its tables are empty, and require it when the source replicates lolor tables. Keep lolor out of create_sub's auto-detected skip_schema so the initial data sync copies the large object data. Add a TAP test covering the rejection cases and end-to-end sync.


ibrarahmad (approved):

ibrarahmad (approved):
LGTM

PR #524, opened 2026-07-10
ACE schema handling

mason-sharp:
ACE keeps node-local state in the pgedge_ace schema using plain Postgres DDL, with no Spock awareness. On a cluster with default AutoDDL settings (spock.enable_ddl_replication=on, spock.include_ddl_repset=on), that breaks both ways:

- ACE's tables get auto-added to a replication set and their contents replicate to peers, or
- once excluded from replication sets, ACE's CREATE TABLE aborts outright on the automatic repset-add, and its DDL text replicates to peers that don't have the schema, failing on apply and (depending on spock.exception_behaviour) disabling the subscription.

A node-local tool must be able to create, modify, and drop its own tables without those statements failing and without anything leaking to peers.

What this PR does

Introduces a built-in never-replicate schema list (builtin_skip_schemas[] = pgedge_ace, exposed via spock_skip_schemas_list() / spock_schema_is_skipped()), enforced at every point where a table can enter replication:

1. AutoDDL — DDL whose identifiable target is a never-replicate schema is not queued for replication, and its tables are not auto-added to any replication set. The statement succeeds locally, silently.
2. Replication set membership — repset_add_table/repset_add_seq refuse relations in such a schema (EnsureRelationNotIgnored), covering repset_add_all_tables and per-partition adds.
3. Indirect repset routing — apply_repset_policy_for_reloid() skips such relations reached indirectly (e.g. a partition under a replicated parent).
4. Structure sync — --exclude-schema is passed unconditionally (pg_dump documents that exclude patterns matching nothing are not an error, and the schema may not exist on the subscriber yet).

No walsender/decode-path changes are needed: because these tables can never enter a replication set, the existing repset gating already suppresses everything — row changes via get_table_replication_info(), TRUNCATE via can_replicate_truncate().

A preparatory commit renames the skip_sys to
nodump_schema[]/nodump_extension[]: tho from the structure-sync dump and do not
prevent replication (lolor is dump-excl the old name collided with the
unrelated per-subscription sub->skip_scvisible names change.

Testing

New TAP test 033_skip_schemas.pl (added to the schedule), 2-node cluster with AutoDDL on:

- CREATE TABLE pgedge_ace.* succeeds an
- repset_add_table on a pgedge_ace tabl
- structure sync excludes pgedge_ace onles still sync)
- DDL against pgedge_ace after the subslicate and does not disable the
subscription (fence-row check)

001_basic, 002_create_subscriber, and 0pass unchanged.

Known limitations (documented in the fe

- A DROP of a never-replicate relation be classified after execution and still
replicates; schema-qualify drops.
- A statement mixing never-replicate an as before.
- Repset memberships that predate this . SET SCHEMA pgedge_ace on a replicated
table — keep replicating until removed SET SCHEMA semantics are tracked
separately.

Out of scope / follow-ups

- User definable custom skip schemas (ex: a future spock.skip_schemas)
- ALTER TABLE ... SET SCHEMA into/out oeparate ticket

### DIFF
--- a/include/spock.h
+++ b/include/spock.h
@@ -117,6 +117,8 @@ extern void gen_slot_name(Name slot_name, char *dbname,
 						  const char *subscriber_name);
 
 extern List *textarray_to_list(ArrayType *textarray);
+extern List *spock_skip_schemas_list(void);
+extern bool spock_schema_is_skipped(const char *nspname);
 extern bool parsePGArray(const char *atext, char ***itemarray, int *nitems);
 
 extern Oid	get_spock_table_oid(const char *table);
--- a/include/spock_node.h
+++ b/include/spock_node.h
@@ -61,8 +61,8 @@ typedef struct SpockSubscription
 } SpockSubscription;
 
 /* NULL-terminated arrays */
-extern const char *const skip_schema[];
-extern const char *const skip_extension[];
+extern const char *const nodump_schema[];
+extern const char *const nodump_extension[];
 
 extern void create_node(SpockNode *node);
 extern void drop_node(Oid nodeid);
--- a/src/spock.c
+++ b/src/spock.c
@@ -858,6 +858,62 @@ spock_supervisor_main(Datum main_arg)
 	proc_exit(0);
 }
 
+/*
+ * Built-in never-replicate schemas, returned by spock_skip_schemas_list().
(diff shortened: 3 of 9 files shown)


mason-sharp:
Used this PR instead: https://github.com/pgEdge/spock/pull/530

PR #527, opened 2026-07-14
Make failover slot worker sync interval configurable.

ibrarahmad:
The spock_failover_slots worker synced slot state to the standby on a hard-coded 60s nap, so the standby's synchronized slots could lag the primary by up to that long, widening the window in which a promotion finds a stale slot. Expose the interval as spock.failover_slots_naptime (and the feedback-wait retry as spock.failover_slots_feedback_naptime), both SIGHUP-settable in milliseconds, so operators can shorten it. Add a TAP test covering the GUC defaults, bounds, and reload.

SPOC-612

### DIFF
--- a/src/spock_failover_slots.c
+++ b/src/spock_failover_slots.c
@@ -61,7 +61,7 @@
 #include "libpq/libpq.h"
 
 /* Defaults (milliseconds) for the tunable worker intervals below. */
-#define WORKER_NAP_TIME_DEFAULT 60000
+#define WORKER_NAP_TIME_DEFAULT 1000
 #define WORKER_WAIT_FEEDBACK_DEFAULT 10000
 
 typedef struct RemoteSlot
--- a/src/spock_failover_slots.c
+++ b/src/spock_failover_slots.c
@@ -60,8 +60,9 @@
 #include "libpq/auth.h"
 #include "libpq/libpq.h"
 
-#define WORKER_NAP_TIME 60000L
-#define WORKER_WAIT_FEEDBACK 10000L
+/* Defaults (milliseconds) for the tunable worker intervals below. */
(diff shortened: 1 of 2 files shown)


src/spock_failover_slots.c
```diff
@@ -1275,9 +1286,9 @@ spock_failover_slots_main(Datum main_arg)
 		 * use long nap so we never elog(ERROR) for hot_standby_feedback off.
 		 */
 		if (RecoveryInProgress() && hot_standby_feedback)
-			sleep_time = synchronize_failover_slots(WORKER_NAP_TIME);
+			sleep_time = synchronize_failover_slots(spock_failover_slots_naptime);
 		else
-			sleep_time = WORKER_NAP_TIME * 10;
+			sleep_time = (long) spock_failover_slots_naptime * 10;
```
mason-sharp, line 1291: If we configure a short nap time, what is the the effect of this * 10?  Is it safe that way, or should it be yet another configuration value? (would be still nice to avoid)

tests/tap/t/026_failover_slots_naptime_guc.pl
```diff
@@ -0,0 +1,96 @@
...
+isnt($rc, 0, "ALTER SYSTEM rejects 10ms (below the 1s minimum)");
+like($out, qr/outside the valid range|invalid value/i,
+     "rejection message mentions the valid range");
+
+# Restore the default so the node is clean for teardown.
+psql_try("ALTER SYSTEM RESET spock.failover_slots_naptime");
+psql_try("SELECT pg_reload_conf()");
+
+destroy_cluster();
+done_testing();
```
mason-sharp, line 96: Doesn't seem like this test helps much. I suppose we could keep it and exclude from the schedule like you have done here.

mason-sharp (approved):

Issue #528, opened 2026-07-14
table_data_filtered() returns corrupt rows for tables with dropped columns (wrong/NULL data, provider segfault)


ilexius:
Seen on spock 5.0.9 and 5.0.10 (`pgedge-postgres:17-spock5-standard`, PG
17.10). The relevant code is unchanged on current main.

## Problem

`spock_table_data_filtered()` (spock_functions.c) builds
`SELECT * FROM <table> [WHERE <row filters>]`, runs it through SPI and puts
the result tuples into the tuplestore as-is:

```c
rc = SPI_execute(query.data, true, 0);
...
for (i = 0; i < SPI_processed; i++)
{
    HeapTuple tup = SPI_tuptable->vals[i];
    tuplestore_puttuple(tupstore, tup);
}
```

The tuplestore descriptor is the relation's tuple descriptor (the function
asserts `equalTupleDescs(tupdesc, reltupdesc)` earlier). But `SELECT *`
omits dropped columns, so SPI materializes its tuples with
natts = live columns only. Consumers then deform those tuples against the
relation descriptor, which still contains the dropped attributes. Every
attribute after the first dropped position is read at the wrong offset:
attributes beyond the tuple's natts come back NULL, attributes within it
return a neighbouring column's datum, and a misread varlena pointer
detoasts into a segfault.

This function feeds the provider-side COPY of every filtered initial sync
and every `sub_resync_table`, so the result is either silently wrong data
on the subscriber (the sync reports success) or a crashed provider
backend. In our case the subscriber's sync worker retried every ~20 s and
kept a production database in crash recovery until we disabled the
subscription.

## Reproduction (silent-NULL variant)

```sql
CREATE TABLE public.tdf_repro (id int PRIMARY KEY, a text, junk text, b text);
INSERT INTO public.tdf_repro SELECT g, 'a'||g, 'x', 'b'||g FROM generate_series(1,100) g;
ALTER TABLE public.tdf_repro DROP COLUMN junk;
SELECT spock.repset_create('tmp_tdf_repro');
SELECT spock.repset_add_table('tmp_tdf_repro','public.tdf_repro', false, row_filter := 'id > 0');

SELECT count(*), max(b)
FROM (SELECT b FROM spock.table_data_filtered(
        NULL::public.tdf_repro, 'public.tdf_repro'::regclass,
        ARRAY['tmp_tdf_repro'])) x;
-- expected: 100 | b99      actual: 100 | NULL
```

[...]

PR #540, opened 2026-07-17
Remap filtered rows to relation descriptor for dropped columns

SAY-5:
`spock_table_data_filtered()` runs `SELECT * FROM <rel>` through SPI and puts the result tuples straight into the tuplestore, whose descriptor is the full relation descriptor. `SELECT *` omits dropped columns, so those tuples only carry the live attributes and every attribute past a dropped position is then read at the wrong offset, giving corrupt/NULL data on the subscriber or a provider crash when a misread varlena is detoasted (#528).

This remaps each SPI tuple to the relation descriptor with `convert_tuples_by_position()` / `execute_attr_map_tuple()`, which inserts NULLs at the dropped positions; when the descriptors already match, `convert_tuples_by_position()` returns NULL and the tuples are stored unchanged.

Reproducing and adding the row-filter regression case both need a running provider/subscriber cluster, which I could not spin up locally, so I verified the change compiles clean against the PG server headers but did not run the sync suite. A minimal SQL repro from the issue:

```sql
CREATE TABLE public.tdf_repro (id int PRIMARY KEY, a text, junk text, b text);
INSERT INTO public.tdf_repro VALUES (1, 'aa', 'xx', 'bb');
ALTER TABLE public.tdf_repro DROP COLUMN junk;
-- before: spock.table_data_filtered() returns b as NULL / misaligned; after: correct rows
```

### DIFF
--- a/src/spock_functions.c
+++ b/src/spock_functions.c
@@ -3040,11 +3040,21 @@ spock_table_data_filtered(PG_FUNCTION_ARGS)
 		for (i = 0; i < SPI_processed; i++)
 		{
 			HeapTuple	tup = SPI_tuptable->vals[i];
+			HeapTuple	remapped = NULL;
 
 			if (map != NULL)
-				tup = execute_attr_map_tuple(tup, map);
+				tup = remapped = execute_attr_map_tuple(tup, map);
 
 			tuplestore_puttuple(tupstore, tup);
+
+			/*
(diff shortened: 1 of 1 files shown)


danolivo (approved):
LGTM.
XXX: It should follow with more brave removal of all that legacy stuff in the future version (fresh vanilla's COPY machinery doesn't have this problem at all). But basically this is a good in-place fix.
NOTE: no tests provided yet.

PR #545, opened 2026-07-24
Add in-core attach_node/detach_node alongside SQL add_node/remove_node

ibrarahmad:
Spock's Zero Downtime Add/Remove Node (ZODAN) workflow lets you grow or shrink a cluster without interrupting the nodes already serving traffic. Until now it lived entirely in SQL scripts that leaned on the `dblink` extension to reach the rest of the cluster. This change brings that same workflow into the extension itself.

Adding a node is now as simple as installing Spock and calling `spock.attach_node` on the new member, and removing one is a single `spock.detach_node` on the node that is leaving. There is nothing to load and no `dblink` to install. The procedures create the node, wire up replication in both directions, synchronize the initial data, line up replication slots and origins, and switch replication on, all from inside Spock and all over the cluster's own connections. Every wait is bounded, so a join that cannot make progress fails fast with a clear error instead of hanging.

The original SQL-script path stays exactly where it was for anyone who prefers to keep the orchestration in a script they can read and adjust, and `spock.add_node` and `spock.remove_node` continue to work unchanged. The two approaches sit side by side under names that never collide, and the Zodan documentation now walks through both in one place.

### DIFF
--- a/docs/modify/zodan/zodan_readme.md
+++ b/docs/modify/zodan/zodan_readme.md
@@ -1,64 +1,123 @@
-# Zodan: Zero-Downtime Node Addition for Spock
+# Using Zodan: Zero-Downtime Node Addition and Removal
 
-Zodan provides tools to add or remove a node with zero downtime. The
-scripts are located in the 
-[samples/Z0DAN](https://github.com/pgEdge/spock/tree/main/samples/Z0DAN)
-directory of the [Spock GitHub](https://github.com/pgEdge/spock)
-repository.
+Zodan adds or removes a node from a Spock cluster with zero downtime for the
+existing nodes. Spock ships two implementations of the same workflow: the
+in-core procedures `spock.attach_node` / `spock.detach_node`, and the SQL
+scripts that provide `spock.add_node` / `spock.remove_node`. Both are covered
+below.
 
-Zodan's workflows and scripts streamline the process of adding a node to
-or removing a node from a Spock cluster. Zodan features the following
-scripts and workflows:
+## In-core procedures: attach_node and detach_node
 
-- The [zodan.sql](#using-the-zodansql-sql-workflow) workflow is a complete
-  SQL-based workflow that uses `dblink` to perform the same add node
-  operations from within Postgres.
-- The [zodremove.sql](#the-zodremovesql-workflow) workflow is a complete
-  SQL-based workflow that uses `dblink` to perform the same removal
-  operations from within Postgres.
+The in-core implementation is built into the Spock extension. A single
+`CREATE EXTENSION spock` makes both procedures available. There is nothing to
+install beyond the extension itself: the `dblink` extension is not required,
+and there are no SQL scripts to load. All local work runs over SPI and all
+cross-node work runs over libpq.
 
-## Components
+- `spock.attach_node` must be run on the new node being added.
+- `spock.detach_node` must be run on the node being removed.
 
-The following scripts and workflows are available via Zodan.
+### Adding a node
 
-### Using the zodan.sql SQL Workflow
-
-The SQL-based implementation utilizes the Postgres `dblink` extension to
-handle node addition directly from within the database. This method is
-ideal for environments where you may not have access to a shell or Python.
-
-Within the workflow, SQL commands orchestrate the following operations:
-
-- `add_node` - The main procedure to orchestrate the full workflow.
-- `create_node` - Register the new node via `spock.node_create`.
-- `get_spock_nodes` - Fetch current node metadata from a remote node.
-- `create_sub` and `enable_sub` - Manage subscription creation and
-  activation.
-- `create_replication_slot` - Create and configure logical replication
-  slots.
-- `sync_event` and `wait_for_sync_event` - Coordinate data synchronization
-  events.
-- `get_commit_timestamp` and `advance_replication_slot` - Align
-  replication states.
-
-To use the workflow, execute the following command in your Postgres
-session.  In the following example, the `spock.add_node` procedure adds a 
-new node to the cluster:
+`spock.attach_node` orchestrates the full join: it validates prerequisites,
+creates the node, sets up replication slots and subscriptions in both
+directions, coordinates synchronization events, advances replication slots and
+origins to a consistent point, and enables replication. It supports both a two
+node cluster and the general multi node case.
 
 ```sql
-CALL spock.add_node(
-  'source_node_name',
-  'src_dsn',
-  'new_node_name',
-  'new_node_dsn',
-  true|false,               -- verbose? optional
-  'new_node_location',      -- optional
-  'new_node_country',       -- optional
-  '{}'::jsonb               -- optional info
+CALL spock.attach_node(
+  src_node_name     => 'source_node_name',
+  src_dsn           => 'src_dsn',
+  new_node_name     => 'new_node_name',
+  new_node_dsn      => 'new_node_dsn',
+  verb              => false,          -- verbose progress output, optional
+  new_node_location => 'NY',           -- optional
+  new_node_country  => 'USA',          -- optional
+  new_node_info     => '{}'::jsonb,    -- optional metadata
+  timeout_sec       => 180             -- bound on each wait, optional
 );
 ```
 
-In the following example, the command adds node `n4` to the cluster:
+In the following example, the command adds node `n4` to the cluster. Run it
+while connected to `n4`:
+
+```sql
+CALL spock.attach_node(
+  'n1',
+  'host=127.0.0.1 dbname=pgedge port=5431 user=pgedge password=<PASSWORD>',
+  'n4',
+  'host=127.0.0.1 dbname=pgedge port=5434 user=pgedge password=<PASSWORD>'
+);
+```
+
+The `timeout_sec` argument bounds every internal wait loop, so a join that
+cannot make progress fails quickly with a clear error rather than blocking for
+a long fixed period. It defaults to 180 seconds; pass a larger value if your
+environment needs more headroom.
+
+### Removing a node
+
+`spock.detach_node` removes a node in the correct order: it drops the inbound
+subscriptions on each surviving node, drops the subscriptions local to the
+node being removed, drops the node's replication sets, and finally drops the
+node from the catalog. Replication slot and origin cleanup is handled as part
+of dropping the subscriptions. A surviving node that cannot be reached during
+teardown is skipped with a warning rather than aborting the removal.
+
+```sql
+CALL spock.detach_node(
+  target_node_name => 'target_node_name',
+  target_node_dsn  => 'target_dsn',    -- DSN of the node being removed
+  verbose_mode     => true             -- verbose progress output, optional
+);
+```
+
+`detach_node` runs on the node being removed and does all of its work locally
+and against the surviving nodes it already knows about. `target_node_dsn` must
+be the DSN of that same node: `detach_node` connects to it and compares the
+system identifier and database name against the local ones, refusing to run if
+they differ. This is the mirror of the `new_node_dsn` check in `attach_node`,
+and it is what stops `detach_node` from tearing down the replication of
+whichever node it was accidentally run on.
+
+In the following example, the command removes node `n4` from the cluster. Run
+it while connected to `n4`:
+
+```sql
+CALL spock.detach_node(
+  'n4',
+  'host=127.0.0.1 dbname=pgedge port=5434 user=pgedge password=<PASSWORD>'
+);
+```
+
+## SQL scripts: add_node and remove_node
+
+The SQL-based implementation uses the Postgres `dblink` extension to run the
+same node add and remove operations from within the database. The scripts are
+located in the
+[samples/Z0DAN](https://github.com/pgEdge/spock/tree/main/samples/Z0DAN)
+directory of the [Spock GitHub](https://github.com/pgEdge/spock) repository.
+This method is useful where you prefer to keep the orchestration in a script
+you can read and modify. Load `zodan.sql` and `zodremove.sql` on the node you
+run the procedures from, and make sure `dblink` is installed there.
+
+### Adding a node with zodan.sql
+
+The `zodan.sql` workflow orchestrates the following operations:
+
+- `add_node` - the main procedure to orchestrate the full workflow.
+- `create_node` - register the new node via `spock.node_create`.
+- `get_spock_nodes` - fetch current node metadata from a remote node.
+- `create_sub` and `enable_sub` - manage subscription creation and activation.
+- `create_replication_slot` - create and configure logical replication slots.
+- `sync_event` and `wait_for_sync_event` - coordinate data synchronization
+  events.
+- `get_commit_timestamp` and `advance_replication_slot` - align replication
+  states.
+
+To use the workflow, load `zodan.sql` and call `spock.add_node`. In the
+following example, the command adds node `n4` to the cluster:
 
 ```sql
 CALL spock.add_node(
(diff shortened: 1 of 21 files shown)


docs/modify/zodan/zodan_readme.md
```diff
@@ -1,110 +1,146 @@
...
+  target_node_dsn  => 'target_dsn',       -- accepted for symmetry, not used
+  verbose_mode     => true             -- verbose progress output, optional
 );
 ```
 
-In the following example, the command adds node `n4` to the cluster:
+`detach_node` runs on the node being removed and does all of its work locally
+and against the surviving nodes it already knows about, so `target_node_dsn`
+is currently accepted for signature symmetry with `spock.remove_node` but is
+not used. Pass any value (for example the node's own DSN).
```
mason-sharp: If target_node_dsn s not used, I think we should remove it.

docs/modify/zodan/zodan_tutorial.md
```diff
@@ -23,19 +23,22 @@ node), `n2`, `n3`, and the new node is `n4`.
...
-      (3-minute) timeout when waiting for sync events between nodes. If
-      replication lag exceeds this window, the procedure will raise an
-      exception. Ensure network latency and replication lag are within
-      acceptable limits before starting. When following the manual workflow
-      below, you call `spock.wait_for_sync_event()` directly and supply your
-      own timeout (the examples use 1200 seconds), so this internal limit
-      does not apply.
+    - The automated `spock.attach_node()` procedure bounds each internal wait
+      (for sync events between nodes) with its `timeout_sec` argument, which
+      defaults to 180 seconds (3 minutes). If replication lag exceeds this
```
mason-sharp, line 35: Do you think we should increase the timeout and also explain that it may take a long time if the user has a lot of data?

samples/Z0DAN/zodan.sql
```diff
@@ -284,13 +284,10 @@ BEGIN
...
         BEGIN
-            -- Query existing schemas on the new node (excluding system schemas).
-            -- lolor is excluded here: Spock already keeps it out of the
-            -- structure dump globally, and putting it into skip_schema would
-            -- also exclude the lolor tables from the initial data sync.
+            -- Query existing schemas on the new node (excluding system schemas)
             remotesql_schema := 'SELECT string_agg(schema_name, '','') as schemas
                                 FROM information_schema.schemata
-                                WHERE schema_name NOT IN (''information_schema'', ''pg_catalog'', ''pg_toast'', ''spock'', ''public'', ''lolor'')
+                                WHERE schema_name NOT IN (''information_schema'', ''pg_catalog'', ''pg_toast'', ''spock'', ''public'')
```
mason-sharp, line 290: If these lolor related changes are needed and not directly related to the in-spock zodan, can you break it out into a separate PR and check if similar changes are needed for v5_STABLE?

sql/spock--6.0.0.sql
```diff
@@ -963,3 +856,34 @@ CREATE FUNCTION spock.reset_subscription_stats(subid oid DEFAULT NULL)
 RETURNS void
 AS 'MODULE_PATHNAME', 'spock_reset_subscription_stats'
 LANGUAGE C CALLED ON NULL INPUT VOLATILE;
+
+-- ----------------------------------------------------------------------------
+-- Zero Downtime Add/Remove Node (ZODAN), in-core C orchestration
+--
```
mason-sharp, line 862: Also add to`spock--5.0.11--6.0.0.sql`

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+ * Progress reporting.  All progress output is gated on verbose mode so a
+ * normal run is quiet; only the final confirmation is emitted unconditionally
+ * by the entry points.  zodan_verbose is set once per top-level call.
+ */
+static bool zodan_verbose = false;
+
+#define ZNOTE(...) \
+	do { if (zodan_verbose) ereport(NOTICE, (errmsg(__VA_ARGS__))); } while (0)
+#define ZVERB(verb, ...) \
+	do { if (zodan_verbose) ereport(NOTICE, (errmsg(__VA_ARGS__))); } while (0)
```
mason-sharp, line 157: These two look the same. Also, can you rename ZVERB to ZVERBOSE to make it clearer. Possibly ZNOTE to ZNOTICE, too. That is, if these should be two different things.

mason-sharp (commented):
I am still reviewing, but here is some initial feedback.

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+{
+	*major = *minor = *patch = 0;
+	if (v == NULL)
+		return;
+	sscanf(v, "%d.%d.%d", major, minor, patch);
+}
+
+/* Return <0, 0, >0 comparing a.b.c to the numbers in the arguments. */
+static int
+zodan_version_cmp(int a1, int a2, int a3, int b1, int b2, int b3)
```
mason-sharp: Can you give these arguments better names?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+static PGconn *zodan_conns[ZODAN_MAX_CONNS];
+static int	zodan_nconns = 0;
+
+/* Connect to a remote node, ERROR on failure.  Tracks the connection. */
+static PGconn *
+zodan_connect(const char *dsn, const char *purpose)
+{
+	PGconn	   *conn = spock_connect(dsn, "spock_zodan", purpose);
+
+	if (conn != NULL && zodan_nconns < ZODAN_MAX_CONNS)
```
mason-sharp: If we hit the max, should we not silently ignore?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+static void
+zodan_check_versions(ZodanAddCtx *ctx)
+{
+	int			min1,
+				min2,
+				min3;
+	int			s1,
+				s2,
+				s3;
+	int			n1,
```
mason-sharp: can you use more descriptive variable names?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+					 quote_literal_cstr(sub_name),
+					 quote_literal_cstr(provider_dsn),
+					 ZODAN_REPSETS,
+					 sync_structure ? "true" : "false",
+					 sync_data ? "true" : "false",
+					 enabled ? "true" : "false");
+
+	if (on_dsn == NULL)
+	{
+		zodan_local_command(sql.data);
```
mason-sharp, line 1303: If local should it just call spock_create_subscription instead?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+
+	ZNOTE("Phase 8: Enabling disabled subscriptions");
+
+	if (zodan_num_other_nodes(ctx) == 0)
+	{
+		char	   *sub_name = zodan_gen_sub_name(ctx->src_node_name, ctx->new_node_name);
+		char	   *lsn;
+
+		zodan_local_command(psprintf(
+			"SELECT spock.sub_enable(subscription_name := %s, immediate := true)",
```
mason-sharp, line 1939: Should it just call spock_alter_subscription_enable()? Here and elsewhere use the local function?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+
+	if (SPI_processed > 0)
+	{
+		bool		isnull;
+		Datum		d = SPI_getbinval(SPI_tuptable->vals[0],
+									  SPI_tuptable->tupdesc, 1, &isnull);
+
+		if (!isnull)
+		{
+			MemoryContext old = MemoryContextSwitchTo(CurrentMemoryContext);
```
mason-sharp: why switch to the current context? Maybe you meant something else.

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+			char	   *s = SPI_getvalue(SPI_tuptable->vals[0],
+										 SPI_tuptable->tupdesc, 1);
+
+			MemoryContextSwitchTo(old);
+			(void) d;
+			ret = s;
+		}
```
mason-sharp: Can this whole block be reduced to:

```
if (SPI_processed > 0)
    ret = SPI_getvalue(SPI_tuptable->vals[0], SPI_tuptable->tupdesc, 1);
```

SPI_getvalue() already returns NULL when the attribute is SQL NULL, if not mistaken

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+				char	   *sub = PQgetvalue(res, j, 0);
+				char	   *en = PQgetvalue(res, j, 1);
+
+				if (en[0] != 't')
+				{
+					PQclear(res);
+					zodan_disconnect(nconn);
+					ereport(ERROR,
+							(errmsg("node %s has disabled subscription %s",
+									ctx->nodes[i].name, sub),
```
mason-sharp, line 1014: Does `sub` point to a location freed by `PQclear`?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+	 * failure to reach or update one node (for example a node that is down) is
+	 * tolerated and does not abort the whole removal, matching the old
+	 * zodremove.sql behavior.  Without the subtransaction, catching the error
+	 * would leave the surrounding transaction in an aborted state.
+	 */
+	for (i = 0; i < nnodes; i++)
+	{
+		MemoryContext subctx = CurrentMemoryContext;
+		ResourceOwner subowner = CurrentResourceOwner;
+		PGconn	   *conn = NULL;
```
mason-sharp: make `conn` volatile

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+	zodan_verbose = verb;
+
+	nonatomic = fcinfo->context != NULL && IsA(fcinfo->context, CallContext) &&
+		!((CallContext *) fcinfo->context)->atomic;
+	if (!nonatomic)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_TRANSACTION_TERMINATION),
+				 errmsg("spock.detach_node() cannot run inside a transaction block"),
+				 errhint("Invoke it with a top-level CALL, not inside BEGIN/COMMIT.")));
+
```
mason-sharp, line 2388: Do we need a PG_TRY - CATCH/FINALLY to clean things up in case there is a problem?

src/spock_zodan.c
```diff
@@ -0,0 +1,1822 @@
...
+ * ereport(ERROR)s mid-orchestration does not strand its socket for the rest of
+ * the backend's life.  spock_connect() does not register the PGconn with any
+ * resource owner, and the phase helpers keep the connection in a local
+ * variable, so on error the entry point drains this registry (see
+ * zodan_close_all_conns).  A handful of slots is far more than the two or three
+ * connections ever open at once.
+ */
+#define ZODAN_MAX_CONNS 16
+static PGconn *zodan_conns[ZODAN_MAX_CONNS];
+static int	zodan_nconns = 0;
```
mason-sharp, line 319: Should this ever be reset somewhere below?

mason-sharp (changes requested):
I did another pass

ibrarahmad:
All valid, fixed in 16246af. Two I did differently: phase 9 now skips the source so phase 10 stays (it reaches the source over the verified src_dsn, not the unverified registered if_dsn), and target_node_dsn is now used for a target identity check rather than dropped.
Also found both procedures missing from spock--5.0.10--6.0.0.sql, so upgrades never got them at all. Added there with the REVOKEs.

docs/modify/zodan/zodan_readme.md
ibrarahmad: Kept it and gave it a job instead: detach_node now connects to target_node_dsn and compares system identifier plus database name against the local ones, refusing to run if they differ. Same check attach_node does on new_node_dsn. Docs updated. 527d9e1

sql/spock--6.0.0.sql
ibrarahmad: Done in 527d9e1. This branch still had the old spock--5.0.10--6.0.0.sql, so I renamed it to spock--5.0.11--6.0.0.sql to match the split on main and added the 5.0.10--5.0.11 stub. Verified 5.0.10 -> 5.0.11 -> 6.0.0 is the only path and both procedures land with owner-only ACL.

src/spock_zodan.c
ibrarahmad: They were the same, ZVERB ignored its first argument entirely and gated on the static like ZNOTE did. Dropped ZVERB and renamed the survivor to ZNOTICE. 527d9e1

src/spock_zodan.c
ibrarahmad: Renamed to left_major/left_minor/left_patch and right_major/right_minor/right_patch. 527d9e1

src/spock_zodan.c
ibrarahmad: Right, silently returning an untracked connection defeated the point of the registry. It now errors out instead. 527d9e1

src/spock_zodan.c
ibrarahmad: Now min_/src_/new_ and node_ prefixed with major, minor, patch spelled out. 527d9e1

src/spock_zodan.c
ibrarahmad: It was a no-op, switching to the context we were already in. Removed along with the unused Datum. 527d9e1

src/spock_zodan.c
ibrarahmad: You are right, SPI_getvalue returns NULL for a SQL NULL. Reduced to exactly that. 527d9e1

src/spock_zodan.c
ibrarahmad: Yes it did, the ereport read it after the result was freed. pstrdup before the clear now. 527d9e1

src/spock_zodan.c
ibrarahmad: Done, with a comment on why. 527d9e1

src/spock_zodan.c
ibrarahmad: Added, detach_node had no drain while attach_node did. Build has --enable-cassert and the assert that both registries are empty on entry never fired across the runs, including repeated failures in one session. 527d9e1

src/spock_zodan.c
ibrarahmad: It is drained in PG_FINALLY in both entry points now, and asserted empty on entry so a leak shows up instead of accumulating. 527d9e1

docs/modify/zodan/zodan_tutorial.md
ibrarahmad: Not changed yet, leaving the call to you. 180s is only the timeout_sec default and the caller can raise it, so my instinct is to document that it scales with data volume rather than raise the default and make a genuinely stuck attach hang longer. Say which you want and I will do it.

samples/Z0DAN/zodan.sql
ibrarahmad: Not done in this push, it needs its own PR. Agreed the samples/Z0DAN lolor bits are separable from the in-core work, and v5_STABLE should be checked for the same gap.

src/spock_zodan.c
ibrarahmad: Deliberate. spock_create_subscription is a PG_FUNCTION_ARGS entry point taking 10 args including three ArrayTypes and an Interval, so calling it directly means hand-building a fcinfo and the arrays. More to the point, this helper feeds the same statement to the local SPI path and the remote libpq path, and the remote one has to be SQL text anyway, so one builder keeps them from drifting. Happy to change it if you would rather not go through SPI locally.

src/spock_zodan.c
ibrarahmad: Same reasoning as the sub_create thread. It is one statement builder shared by the local and remote paths. Can switch the local calls to the C functions if you prefer, it just splits the two paths.

samples/Z0DAN/zodan.sql
```diff
@@ -302,13 +302,10 @@ BEGIN
...
```
mason-sharp: If we need to change handling in zodan.sql, can you break that out as a separate PR that we can review and do first?

tests/tap/t/033_zodan_lolor_add_node.pl
```diff
@@ -98,19 +99,14 @@ sub dsn {
...
 
-# Load the zodan procedures on the node being added.
-system_or_bail("$PG/psql", '-X', '-p', $cfg->{node_ports}[2], '-d', $DB,
-               '-v', 'ON_ERROR_STOP=1', '-f', '../../samples/Z0DAN/zodan.sql');
-pass('zodan procedures loaded on n3');
-
-my $add_node_sql =
-    "CALL spock.add_node('n1', '" . dsn(1) . "', 'n3', '" . dsn(3) . "', " .
+my $attach_node_sql =
+    "CALL spock.attach_node('n1', '" . dsn(1) . "', 'n3', '" . dsn(3) . "', " .
```
mason-sharp: If these two coexist for a while, I think we should keep two versions of this test.

samples/Z0DAN/zodan.sql
ibrarahmad: new PR
https://github.com/pgEdge/spock/pull/598

PR #547, opened 2026-07-27
spock_apply: advance forwarded origin using pre-created disabled subscription

rasifr:
Any node that subscribes with forward_origins='all' receives transactions
    originally sourced from other peer nodes, forwarded through its immediate
    provider. Each forwarded transaction's ORIGIN wire message carries the
    original peer's Spock node OID, not the immediate provider's. A peer's
    transactions can reach this node forwarded this way well before -- or
    without -- a direct subscription to that peer ever being created here. If
    one has been (or later is) created, tracking the forwarded progress on its
    origin lets it start from the right position whenever it is enabled,
    instead of a full resync. A replication origin only exists where a
    subscription entry does, so this requires resolving the peer's node id to
    the local subscription (if any) that names it as provider.

    handle_origin() resolves the peer node ID to a matching local subscription,
    excluding invalid origins and the direct provider, and stores the result in
    forwarded_local_origin_id for the duration of the transaction. This is
    resolved fresh for every ORIGIN message rather than cached across
    transactions: a cached RepOriginId can be silently reassigned to an
    unrelated origin once its owning subscription is dropped (RepOriginId is a
    small, actively recycled space -- replorigin_create() always reuses the
    lowest free id, and replorigin_advance() does no catalog validation against
    it), and Spock's resolution has no cheap way to revalidate a stale value
    later the way a simple name lookup would. Resolving within the same
    transaction that uses it bounds that exposure to a single transaction's
    replay. Multiple matching subscriptions are rejected as ambiguous.

    At commit, maybe_advance_forwarded_origin() advances the resolved origin using
    XactLastCommitEnd as this node's local position, rather than the provider's
    end_lsn from a different WAL space. The WAL-logged origin becomes the durable
    resume position. Because the advance is logged separately after the data
    commit, a crash may replay the latest transaction, preserving at-least-once
    semantics; delta-apply idempotency remains a separate follow-up.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -102,12 +102,12 @@ static TimeOffset apply_delay = 0;
 static TimestampTz required_commit_ts = 0;
 
 /*
- * Cache for forwarded origin lookup. The remote_origin_id (Spock node ID)
- * is consistent across the cluster, so we can use it as a cache key to
- * avoid repeated slot name generation and origin lookups.
+ * Local origin for the current forwarded transaction's peer, if any.
+ * Resolved fresh per ORIGIN message (handle_origin()), not cached --
+ * a cached RepOriginId can be silently reassigned to an unrelated
+ * origin once its subscription is dropped.
  */
-static RepOriginId cached_forward_remote_id = InvalidRepOriginId;
-static RepOriginId cached_forward_local_id = InvalidRepOriginId;
+static RepOriginId forwarded_local_origin_id = InvalidRepOriginId;
 
 static Oid	QueueRelid = InvalidOid;
 
@@ -274,7 +274,9 @@ static void append_feedback_position(XLogRecPtr local_commit_lsn,
 static void get_feedback_position(XLogRecPtr *recvpos, XLogRecPtr *writepos,
 								  XLogRecPtr *flushpos, XLogRecPtr *max_recvpos);
 static void UpdateWorkerStats(XLogRecPtr last_received, XLogRecPtr last_inserted);
-static void maybe_advance_forwarded_origin(XLogRecPtr end_lsn, bool xact_had_exception);
+static RepOriginId resolve_forward_peer_origin(RepOriginId remote_origin_id);
+static void resolve_forwarded_origin_for_transaction(RepOriginId remote_origin_id);
+static void maybe_advance_forwarded_origin(XLogRecPtr local_lsn, bool xact_had_exception);
 static ApplyReplayEntry *apply_replay_queue_next_entry(void);
 static bool apply_replay_queue_append_entry(ApplyReplayEntry **entry_p,
 							StringInfo *msg_p);
@@ -836,6 +838,7 @@ handle_commit(StringInfo s)
 	XLogRecPtr	end_lsn;
 	TimestampTz commit_time;
 	XLogRecPtr	remote_insert_lsn;
+	XLogRecPtr	local_commit_lsn = InvalidXLogRecPtr;
 
 	errcallback_arg.action_name = "COMMIT";
 	xact_action_counter++;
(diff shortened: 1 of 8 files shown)


src/spock_functions.c
```diff
@@ -461,6 +461,67 @@ spock_alter_node_drop_interface(PG_FUNCTION_ARGS)
...
+ * pending state is passed explicitly because its catalog state may not yet
+ * reflect the change. The check covers all subscriptions because forwarding
+ * may relay any origin.
+ *
+ * This catalog check cannot prevent worker races or direct catalog changes.
+ * maybe_advance_forwarded_origin() leaves ownership errors uncaught so they
+ * remain visible.
+ */
+static void
+enforce_forwarding_exclusivity(Oid target_sub_id, const char *target_sub_name,
```
ibrarahmad, line 483: Blocks forward_origins whenever another subscription on this node is enabled (and blocks enabling a plain sub while one forwards). It is the right guard against the ERRCODE_OBJECT_IN_USE advance failure, but it changes behavior for anyone using forward_origins in a live mesh. Intended as join/catchup only? Worth a release note and doc mention.

src/spock_functions.c
```diff
@@ -1029,7 +1097,10 @@ spock_alter_subscription_options(PG_FUNCTION_ARGS)
 			}
 
 			if (strcmp(key, "forward_origins") == 0)
+			{
+				enforce_forwarding_exclusivity(sub->id, sub->name, sub->enabled, result);
```
ibrarahmad, line 1105: On a disabled sub, target_forwards is false, so if another enabled sub forwards this still errors with "cannot enable subscription ..." even though we are only setting forward_origins, not enabling. Consider skipping the check when sub->enabled is false.

src/spock_functions.c
mason-sharp, line 1093: Confirming, only "all" is supported at this time?

rasifr: yes. only "all" is supported.

src/spock_apply.c
```diff
@@ -1175,6 +1181,32 @@ handle_commit(StringInfo s)
...
+resolve_forward_peer_origin(RepOriginId remote_origin_id)
+{
+	List	   *subs;
+
+	subs = get_node_subscriptions((Oid) remote_origin_id, true);
+	if (subs == NIL)
+		return InvalidRepOriginId;
+
+	if (list_length(subs) > 1)
+		elog(ERROR, "SPOCK %s: ambiguous forwarded origin: more than one "
```
ibrarahmad: sub_create only warns on overlapping repsets, it never blocks a second subscription to the same provider node, so this is reachable. Once it fires the forwarding worker throws on every forwarded txn from that peer and never makes progress. A WARNING plus skipping the advance would keep replication alive.

src/spock_apply.c
```diff
@@ -1211,6 +1243,32 @@ handle_origin(StringInfo s)
...
+static void
+resolve_forwarded_origin_for_transaction(RepOriginId remote_origin_id)
+{
+	forwarded_local_origin_id = InvalidRepOriginId;
+
+	if (remote_origin_id == InvalidRepOriginId ||
+		remote_origin_id == (RepOriginId) MySubscription->origin->id)
+		return;
+
+	StartTransactionCommand();
```
ibrarahmad, line 1275: This opens a transaction and scans spk_subscription for every ORIGIN message, so it runs on each forwarded txn on the cascade path where the old code used a cache. Better to resolve once per peer and revalidate than to do a full lookup each time.

rasifr: No. Thats intentional; there are too many caveats with adding cache and had been noted as a future todo; For now this cost is acceptable. Also It only occurs for origin forwarding.

ibrarahmad (approved):

PR #549, opened 2026-07-28
Add in-repo release workflow and spock50 RPM/DEB packaging

maqeel75:
## Summary

  Adds an in-repo release pipeline for the v5 series, so Spock's RPMs and DEBs are
  built from this repo on a tag push instead of out of `pgedge-enterprise-packages`.
  Packaging files are taken from that repo's `spock50/` component; the workflow
  follows the `pgedge-postgres-mcp` / `pgedge-rag-server` shape.

  `.github/workflows/release.yml` — 7 jobs:
  `detect-matrix → determine-repo-type → package-rpm/package-deb → push-dnf/push-apt → publish-manifest`
  across almalinux 9/10 and jammy/noble/resolute/bullseye/bookworm/trixie, amd64 + arm64.

  ## Notable choices

  - **Per-PG fan-out.** `PER_PG_VERSION=true`, but this repo has no `postgresql-N/`
    dirs, so the PG list can't be auto-discovered — passed explicitly, default
    `16,17,18`, overridable per run. Artifact names carry the PG major so the three
    PG cells of one (os, arch) don't collide.
  - **`COMPONENT_NAME=spock50`**, not the repo name, so v5 and v6 never share an S3
    backup path or `_pending/` manifest key.
  - **Source from the checkout**, not a clone: each cell `git archive`s into
    `release-artifacts/`, which the build scripts prefer over cloning the tag.
    Packages are the exact commit under test, and `simulate_tag` branch runs work.
  - **Guards:** tag filter `v5.*` plus an assert that `include/spock.h`
    `SPOCK_VERSION` equals the tag version and is `5.x`.
  - **DEB pre-release ordering fixed** (spock50 lacked it): `5.0.11~rc1-1.noble`
    now sorts below `5.0.11-1.noble`.
  - **Dropped `debian/patches/`** — `pkglibdir.patch` targets a
    `spock_create_subscriber` recipe that no longer exists and does not apply.


danolivo (approved):
The same as for PR#550. Basically, looks good

PR #553, opened 2026-07-31
Introduce PostgreSQL 19 support to Spock v5_STABLE

danolivo:
Ports pgEdge/spock#518 (PostgreSQL 19 support for main) onto the v5_STABLE line.
### Cherry-picked from #518
* **src/compat/19** — PG19 compatibility layer (index_beginscan flags, ShmemInitHash, ReplOriginId/replorigin_xact_state, StaticAssertVariableIsOfType, query_uses_temp_object, etc.)
* **Source adaptations** — TupleDescFinalize() for manually built descriptors, RepackStmt (CLUSTER folded into REPACK), SignalRecoveryConflictWithVirtualXID, const JumbleState in post_parse_analyze hooks, SpockTxnCommitTime() accessor for the flattened ReorderBufferTXN commit time, wait_event.h / lsyscache.h include relocations.
* **CI** — PG19 added to the pg-stable-test, spockbench, and zodan_sync matrices.
* **Docs** — supported-version lists and a release-note entry.
Main-only files touched by #518 (spock_conflict_stat.c, spock_progress_recovery.c, 015_skip_lsn.pl and the installcheck/nightly workflows) do not exist on v5_STABLE and were dropped from the pick.
### v5-specific adaptations (not in #518)
Code that main refactored away or compiles out on PG18+ still exists on v5 and needed its own PG19 fixes:
* `spock_apply_heap.c`: ExecInsertIndexTuples() now takes a flags bitmask with reordered arguments.
* `spock_failover_slots.c`: ReplicationSlotCreate() grew a "repack" argument; ReplicationSlot.active_pid became active_proc (new SlotIsActive() wrapper); the local get_database_oid() copy is renamed to spock_get_database_oid() to avoid the core export PG19 added to catalog/pg_database.h.
* `spock.c`: log_min_messages became a per-backend-type array; the emit-log hook now indexes it with MyBackendType. (Note: main has the same comparison and currently compiles it as a pointer/int comparison on PG19 — separate fix needed there.)
* `spock_output.c`: include spock_compat.h for the AssertVariableIsOfType rename.
* `compat/19`: restore the SPKstandard_ProcessUtility / SPKnext_ProcessUtility_hook macros still used by v5's spock_executor.c.
* `sql/spock--5.0.0.sql`: accept major version 19 in the CREATE EXTENSION version gate.


docs/spock_release_notes.md
```diff
@@ -3,6 +3,11 @@
 ## Spock 5.0.11
 
 ### New Features
+* **PostgreSQL 19 support** — new `compat/19` layer and version-specific API
```
mason-sharp, line 6: Let's move this to a new "Unreleased" section instead of 5.0.11

mason-sharp (changes requested):

mason-sharp (approved):

PR #557, opened 2026-08-04
Cherry-pick PG19-adaptation commits and fixes from v5_STABLE

danolivo:
### Fixes

- **Exception replay mode leaked onto healthy transactions.** `use_try_block` lives in the worker's shmem slot, not per-transaction state. An error raised *between* remote transactions never reaches the commit that clears it, and `apply_work()` enters replay mode anyway — so the flag landed on the next transaction from the stream. `handle_begin()` only reasoned about the flag when `commit_lsn` matched the recorded failure, so a non-matching transaction inherited it, was replayed read-only, and under `TRANSDISCARD`/`SUB_DISABLE` was discarded. A transaction that never failed, lost, and blamed on an unrelated error. Now the flag is derived from the comparison, and the recorded root cause is dropped with it.

- **Off-by-one scan of the exception log array.** Both search loops ran to `i <= SpockCtx->total_workers`, one past the end. `[total_workers]` is outside the region startup zeroes, and an empty-looking name there would be taken as the free slot, so `namestrcpy()` wrote past the array. It landed in slack only because `worker_shmem_size()` over-allocates.

- **Slot name matched by prefix instead of exactly.** `strncmp()` bounded by `strlen(MySubscription->name)` made `test_subscription` share a slot with `test_subscription_parallel` — and sharing the slot means sharing `commit_lsn`, so one subscription's failure could push the other into exception handling for a transaction that never failed. Now `namestrcmp()`.

- **Failover-slots worker died while the walreceiver reconnected.** `WalRcv->conninfo` is empty for as long as a connection attempt is in flight, which is every walreceiver start and restart, not just standby startup — the worker died in that window for 60s at a time. `make_sync_failover_slots_dsn()` now reports whether a DSN is available instead of building an unusable one; both callers wait for the next cycle. The field is also copied under `WalRcv->mutex`, matching core's `pg_stat_get_wal_receiver()`; it was read unlocked while the walreceiver rewrote it under that spinlock.

- **Log-level re-check was inverted.** `spock_error_callback()` downgrades two serialization-failure retry messages to `DEBUG1`, then re-tests whether they should still be emitted — backwards in both comparisons. `DEBUG1` is 14 and the default `log_min_messages` is `WARNING` = 19, so `19 < 14` never fired and the messages went to the server log regardless of the threshold, which is what this code exists to prevent. Now compares in the direction `is_log_level_output()` uses, as `spock_dependency.c` already does.

- **Four PG19 compiler warnings.** `log_min_messages` became an array indexed by backend type (pointer-vs-int comparison, the bug above); `pg_fallthrough` is new in PG19's `c.h` and clang won't accept a comment as the annotation; `hash_get_num_entries()` now returns `int64`, no longer matching `%lu`; `generate_subscription_id()` had an empty parameter list. New compat macros for the 15–18 headers.

### Refactor

- `clear_exception_log_entry()` collects the four assignments that reset a shared exception log entry. Two call sites need the same set, and `local_tuple` is easy to overlook — it's the only non-scalar, and it dangles by the time either caller runs. No functional change.

### DIFF
--- a/src/compat/15/spock_compat.h
+++ b/src/compat/15/spock_compat.h
@@ -110,5 +110,25 @@
 							restart_seqs, run_as_table_owner) \
 		ExecuteTruncateGuts(explicit_rels, relids, relids_logged, behavior, \
 							restart_seqs)
+/*
+ * pg_fallthrough is new in PG19, where c.h picks whichever spelling the
+ * compiler understands.  Provide it for older majors so a deliberate
+ * fall-through between switch labels can be annotated once, portably.  clang
+ * does not accept a plain comment as the annotation the way gcc does, so
+ * without this the switch in handle_queued_message() warns.
+ */
+#ifndef pg_fallthrough
+#if defined(__GNUC__) || defined(__clang__)
+#define pg_fallthrough	__attribute__((fallthrough))
+#else
+#define pg_fallthrough
+#endif
+#endif
+
+/*
+ * PG19 turned log_min_messages into an array indexed by backend type, so that
+ * the threshold can be set per backend type.  Older majors have a single int.
+ */
+#define SPOCK_LOG_MIN_MESSAGES	(log_min_messages)
 
 #endif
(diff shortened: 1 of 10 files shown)


mason-sharp:
@danolivo The first 3 commits look like legitimate cherry-picks. The other ones don't and may possibly be needed in v5_STABLE as well. Please break this out into more than one PR to make tracking these easier.

PR #574, opened 2026-08-12
docs: document output_plugin_libraries requirement in README

zaidshabbir25:
PostgreSQL 15.19, 16.15, 17.11, 18.6, and 19 require logical decoding output plugins to be named in the new output_plugin_libraries parameter before they can be used. Slot creation with Spock's spock_output plugin fails with 'library "spock_output" may not be used as an output plugin' unless it is added to that list.

Add the setting to the postgresql.conf example block and explain that the value must be additive, since replacing the built-in default of 'pgoutput, test_decoding' disables the built-in plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,8 +196,6 @@ Modify the `postgresql.conf` file, adding:
 
 PostgreSQL 15.19, 16.15, 17.11, 18.6, and 19 (and their later minor releases) require every logical decoding output plugin to be named in the `output_plugin_libraries` parameter before it can be used. Spock decodes with the `spock_output` plugin, so `spock_output` must appear in this list on every node, or slot creation fails with `library "spock_output" may not be used as an output plugin`.
 
-The value must be *additive*: the built-in default is `'pgoutput, test_decoding'`, so keep those entries alongside `spock_output`. Setting the parameter to `spock_output` alone disables the built-in plugins.
-
 You'll also want to enable automatic ddl replication on each node; add these GUCs to the `postgresql.conf` file as well:
 
     spock.enable_ddl_replication=on
--- a/README.md
+++ b/README.md
@@ -191,6 +191,12 @@ Modify the `postgresql.conf` file, adding:
     max_wal_senders = 10        # one per node needed on provider node
     shared_preload_libraries = 'spock'
     track_commit_timestamp = on # needed for conflict resolution
+    output_plugin_libraries = 'pgoutput, test_decoding, spock_output'
+                                # required on newer Postgres versions; must be additive
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,6 +191,12 @@ Modify the `postgresql.conf` file, adding:
     max_wal_senders = 10        # one per node needed on provider node
     shared_preload_libraries = 'spock'
     track_commit_timestamp = on # needed for conflict resolution
+    output_plugin_libraries = 'pgoutput, test_decoding, spock_output'
+                                # required on newer Postgres versions; must be additive
+
+PostgreSQL 15.19, 16.15, 17.11, 18.6, and 19 (and their later minor releases) require every logical decoding output plugin to be named in the `output_plugin_libraries` parameter before it can be used. Spock decodes with the `spock_output` plugin, so `spock_output` must appear in this list on every node, or slot creation fails with `library "spock_output" may not be used as an output plugin`.
+
+The value must be *additive*: the built-in default is `'pgoutput, test_decoding'`, so keep those entries alongside `spock_output`. Setting the parameter to `spock_output` alone disables the built-in plugins.
```
danolivo: Why **must** it be additive? It seems to me that:
1. User might decide which replication plugins are allowed in the system.
2. Spock can't be used in parallel with `pgoutput` in the same instance - this option is a nice way to satisfy the restriction at a high level of abstraction.

danolivo (changes requested):

README.md
zaidshabbir25: We can remove the 2nd part only write "PostgreSQL 15.19, 16.15, 17.11, 18.6, and 19 (and their later minor releases) require every logical decoding output plugin to be named in the `output_plugin_libraries` parameter before it can be used. Spock decodes with the `spock_output` plugin, so `spock_output` must appear in this list on every node, or slot creation fails with `library "spock_output" may not be used as an output plugin`."

zaidshabbir25: Removed the 2nd paragraph.

zaidshabbir25:
Change already address under v5_STABLE
"7. Then, update your Postgres postgresql.conf file, setting:

shared_preload_libraries = 'spock'
track_commit_timestamp = on # needed for conflict resolution
output_plugin_libraries = 'pgoutput, test_decoding, spock_output'
Set output_plugin_libraries only if your server has that parameter: it was added by a 2026 security fix (CVE-2026-6471), and unless spock_output is listed there logical decoding fails with library "spock_output" may not be used as an output plugin — on established clusters as well as new ones, because the check runs every time decoding starts. On a release predating the fix the parameter does not exist and setting it stops the server from starting; check with SELECT current_setting('output_plugin_libraries', true).

PR #577, opened 2026-08-13
Table sync with merge when the target already holds rows

ibrarahmad:
A COPY into a populated table aborts on the first duplicate key and leaves the sync status failed, after which apply silently discards every later change for that table.  Stage the load and merge, and warn on the failure.

### DIFF
--- a/src/spock_sync.c
+++ b/src/spock_sync.c
@@ -72,6 +72,16 @@
 #define PGDUMP_BINARY "pg_dump"
 #define PGRESTORE_BINARY "pg_restore"
 
+/*
+ * Staging table used by copy_table_data() when the target already holds rows.
+ * Lives in pg_temp on the target connection for the duration of the COPY.
+ */
+#define SPOCK_SYNC_STAGE_RELNAME	"spock_sync_stage"
+
+/* Savepoint and bound for the empty-table lock in copy_table_data(). */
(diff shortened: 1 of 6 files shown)


mason-sharp:
@ibrarahmad For the original lolor issue, I am wondering if https://github.com/pgEdge/spock/pull/559 will actually address the issue since lolor should be in the reciprocal repsets. This branch could use a rebase to get the latest code.

Still, we may have some general issue with resyncing that we need to address

danolivo (commented):
Reading this and the initial issue, I failed to find a connection. In the issue, there was n3 - a clear node without any LO objects. So, I'd see the reproduction that proves the issue origin was detected correctly.

src/spock_sync.c
```diff
@@ -987,6 +997,67 @@ make_copy_attnamelist(SpockRelation *rel)
...
+	has_rows = PQntuples(res) > 0;
+	PQclear(res);
+
+	return has_rows;
+}
+
+/*
+ * Run one command on the target connection during a table copy.
+ *
+ * Never returns on failure. On success the caller owns the result.
```
mason-sharp: What does never returns on failure mean?

mason-sharp:
Rebased due to conflicts

PR #581, opened 2026-08-17
Add sample script to backfill FAILOVER on existing Spock slots.

ibrarahmad:
Spock sets the FAILOVER flag only at slot creation. Clusters whose subscriptions predate spock.use_native_failover_slots therefore keep unflagged slots after the GUC is turned on, and PostgreSQL's slotsync worker ignores them.

enable_failover_slots.sh flips those slots with ALTER_REPLICATION_SLOT. restart_lsn and confirmed_flush_lsn are preserved, so no resync is needed. It checks every node and slot before changing anything, and has a --dry-run mode.

Also move set_synchronized_standby_slots.sh into samples/failover/ and update samples/README.md.


mason-sharp:
We won't add to the repo after all.

PR #585, opened 2026-08-18
Add test for verifying exception handling

mason-sharp:
For in between transactions.

This needs PR #584 to be merged first before the test passes, but the test is also purposely not in the CI schedule.

### DIFF
--- a/tests/tap/t/037_no_replay_mode_leak.pl
+++ b/tests/tap/t/037_no_replay_mode_leak.pl
@@ -1,0 +1,165 @@
+use strict;
+use warnings;
+use Test::More;
+use lib '.';
+use SpockTest qw(
+    create_cluster destroy_cluster
+    get_test_config scalar_query psql_or_bail
+    wait_for_sub_status
+);
+
+# An ERROR raised in the apply worker BETWEEN two remote transactions sends
+# apply_work() into replay mode with an empty replay queue -- handle_commit()
+# reset the queue when the last transaction committed, so there is nothing
+# queued to replay.  MyApplyWorker->use_try_block lives in the worker's shared
+# memory slot rather than in per-transaction state, so it then belongs to
+# whichever transaction the provider sends next: one that has never been
+# attempted, let alone failed.
+#
+# handle_begin() is where that has to be caught.  The exception log entry
+# carries the commit_lsn of the recorded failure, and an incoming transaction
+# whose commit_lsn does not match it is not that failure, so replay mode must
+# be turned off and the entry dropped.  Left on under transdiscard or
+# sub_disable, every action of the healthy transaction runs in a subtransaction
+# that is rolled back unconditionally: the pass applies nothing, and what the
+# operator sees is the exception policy firing on data that never failed.
+#
+# pg_cancel_backend() is what makes the window reachable from SQL.  An idle
+# apply worker sits in WaitLatchOrSocket() followed by CHECK_FOR_INTERRUPTS(),
+# which is between remote transactions by construction, and the cancel raises
+# 57014 there exactly once -- ProcessInterrupts() clears QueryCancelPending
+# before it throws.  57014 is not one of the transient SQLSTATEs covered by
+# 035_deadlock_retry.pl, so it takes the replay path rather than the
+# rethrow-and-restart path, which is the whole point: the worker stays up,
+# carrying replay mode into the next transaction.
+
+create_cluster(2, 'Create 2-node replay mode leak cluster');
+
+my $config = get_test_config();
+my $p1 = $config->{node_ports}->[0];
+my $p2 = $config->{node_ports}->[1];
+my $conn = "host=$config->{host} dbname=$config->{db_name} port=$p1 " .
+           "user=$config->{db_user} password=$config->{db_password}";
+my $subscriber_log = "$config->{log_dir}/00${p2}.log";
+
+# Both modes discard the whole transaction on exception, so both turn a
+# leaked replay pass into a transaction the subscriber never applies.  discard
+# is not covered: it commits each successful action, so a healthy transaction
+# replayed under it still lands and there is nothing to observe.
+my @modes = (
+    { mode => 'transdiscard', table => 'leak_transdiscard' },
+    { mode => 'sub_disable',  table => 'leak_sub_disable' },
+);
+
+# Created before sub_create so that no DDL is replicated during the test.
+for my $phase (@modes) {
+    psql_or_bail(1, "CREATE TABLE $phase->{table} (id int PRIMARY KEY, val text)");
+    psql_or_bail(2, "CREATE TABLE $phase->{table} (id int PRIMARY KEY, val text)");
+}
+
+psql_or_bail(2,
+    "SELECT spock.sub_create('sub_n1_n2', '$conn', " .
+    "ARRAY['default', 'default_insert_only', 'ddl_sql'], false, false)");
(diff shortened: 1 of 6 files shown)


tests/tap/t/037_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
...
+    # compared against after the cancel.
+    psql_or_bail(1, "INSERT INTO $t VALUES (1, 'before cancel')");
+    is(wait_for_row($t, 1), '1', "$mode: baseline transaction replicates");
+
+    my $pid_before = apply_worker_pid();
+    like($pid_before, qr/^\d+$/, "$mode: found the apply worker ($pid_before)");
+
+    # The worker is idle here, so the cancel lands between transactions with
+    # nothing queued -- the state the fix is about.
+    sleep(2);
```
danolivo: I think a sleep call here is a source of instability. I'd recommend using sync_event instead.

tests/tap/t/037_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
...
+
+    psql_or_bail(2, "ALTER SYSTEM SET spock.exception_behaviour = $mode");
+    psql_or_bail(2, "SELECT pg_reload_conf()");
+    sleep(2);
+
+    # A clean transaction first: it is what leaves a commit_lsn behind in the
+    # worker's exception log slot, which is what the incoming transaction is
+    # compared against after the cancel.
+    psql_or_bail(1, "INSERT INTO $t VALUES (1, 'before cancel')");
+    is(wait_for_row($t, 1), '1', "$mode: baseline transaction replicates");
```
danolivo: I'd say it is too complicated. Use existing machinery: sync_event and table selection instead.

tests/tap/t/037_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
...
+    return scalar_query(2, "SELECT count(*) FROM $table WHERE id = $id");
+}
+
+for my $phase (@modes) {
+    my $t    = $phase->{table};
+    my $mode = $phase->{mode};
+
+    psql_or_bail(2, "ALTER SYSTEM SET spock.exception_behaviour = $mode");
+    psql_or_bail(2, "SELECT pg_reload_conf()");
+    sleep(2);
```
danolivo: That's a little annoying and delays tests. Maybe we can introduce a routine to apply extra parameters at instance start? It might help other tests to be a little more stable (and faster).

tests/tap/t/037_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
...
+    like($pid_before, qr/^\d+$/, "$mode: found the apply worker ($pid_before)");
+
+    # The worker is idle here, so the cancel lands between transactions with
+    # nothing queued -- the state the fix is about.
+    sleep(2);
+    my $log_offset = -s $subscriber_log // 0;
+    is(scalar_query(2, "SELECT pg_cancel_backend($pid_before)"), 't',
+        "$mode: cancel signal sent to the apply worker");
+
+    ok(wait_for_log($log_offset,
```
danolivo: Maybe just wait until the current backend is removed from the activity view?

tests/tap/t/039_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
+use strict;
```
danolivo, line 1: Check if 037 is free to use in both main and v5 branches

tests/tap/t/039_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,165 @@
...
+       '0', "$mode: nothing is logged to exception_log");
+    is(scalar_query(2,
+           "SELECT sub_enabled FROM spock.subscription WHERE sub_name = 'sub_n1_n2'"),
+       't', "$mode: subscription is not disabled");
+    ok(wait_for_sub_status(2, 'sub_n1_n2', 'replicating', 30),
+        "$mode: subscription stays in replicating state");
+}
+
+destroy_cluster('Destroy replay mode leak cluster');
+done_testing();
```
danolivo, line 128: Basically, nice sketch. Just make it simpler and introduce general mechanics into SpockTest.pm if needed - especially if vanilla's one already has something similar.

tests/tap/t/039_no_replay_mode_leak.pl
mason-sharp: changed to 039

tests/tap/t/039_no_replay_mode_leak.pl
mason-sharp: I added some helpers. I almost made it a separate PR, but kept it here.

tests/tap/t/037_no_replay_mode_leak.pl
mason-sharp: For this test the behavior changes so I did not change here, but feel free to do it as a separate PR.

tests/tap/t/039_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,128 @@
...
+
+    # The transaction that must not be treated as the recorded failure.
+    psql_or_bail(1, "INSERT INTO $t VALUES (2, 'after cancel')");
+    ok(sync_nodes(1, 2),
+        "$mode: worker consumes the stream past the next transaction");
+
+    is(scalar_query(2, "SELECT count(*) FROM $t WHERE id = 2"), '1',
+        "$mode: the next transaction applies");
+
+    unlike(log_since(2, $offset),
```
danolivo, line 112: log_since() returns '' in case if the open() fails - so unsuccessful operation passes successfully, right?

tests/tap/t/SpockTest.pm
```diff
@@ -489,14 +509,105 @@ sub wait_for_pg_ready {
...
     }
     return 0;
 }
 
+# Path to a node's server log.  create_postgresql_conf() names it after the
+# port, so this has to agree with the log_filename set there.
+sub node_logfile {
+    my ($node_num) = @_;
+    my $port = ($BASE_PORT + $node_num - 1);
+    return "$LOG_DIR/00$port.log";
```
danolivo, line 533: This is a problem. It is based on $TESTLOGDIR, but without cherry-picking the commit c45f722, it is initialised only by run_tests.sh. So, we can't run this TAP test individually.

tests/tap/t/SpockTest.pm
```diff
@@ -489,14 +509,105 @@ sub wait_for_pg_ready {
...
+}
+
+# PID of the apply worker on $node_num, or '' if none appears before timeout.
+# spock copies the bgworker name into application_name, so the worker shows up
+# as "spock apply <db>:<sub>".
+sub apply_worker_pid {
+    my ($node_num, $timeout) = @_;
+    $timeout //= 30;
+    my $query = "SELECT pid FROM pg_stat_activity " .
+                "WHERE application_name LIKE 'spock apply %'";
```
danolivo: Is there any chance of getting two records? - parallel subscription, different databases, whatever? I'd use LIMIT or/and datname filter to be sure

danolivo (changes requested):

danolivo (approved):
There are some places where potentially false positives might occur - late SIGHUP or delayed log flush, for example. But nothing that actually blocks the merge itself.

PR #595, opened 2026-08-24
Nightly cross-major cluster upgrade

danolivo:
Adds a nightly job that answers one question: does a Spock cluster survive a cross-major PostgreSQL upgrade?

```md
NOTE: for the sake of a comprehensive check, it was almost entirely generated and tested by an AI agent. As a result, successful passes should be treated sceptically until a couple of manual correction cycles have passed. The primary interest here is negative passes that unlock fast detection of broken compatibility.
```

It runs `tests/run-cluster-upgrade.sh`, which builds everything it needs and does the upgrade the way it would be done in production — a node at a time:

- build two PostgreSQL majors, plus a Spock for each (`v5_STABLE` for the pre-upgrade side, this branch for the upgraded side);
- bring up a three-node mesh on the older major and populate it with the core PostgreSQL regression database (~700 relations: every object kind, inheritance, partitioning, matviews, foreign tables, C functions, odd types, dropped columns, rewritten heaps);
- join the remaining nodes with `spock.add_node()`;
- then, per node: fence it, drain its slots to a sync-event LSN, record its inbound positions, `pg_upgrade`, restore the positions, bring the peers' subscription back, and verify before moving to the next one.

Verification is not just "the subscriptions say replicating". A witness table is written **on the peers while each node is away**, so a change dropped during the window is a missing row rather than something nothing looks at; a `spock.sync_event()` is driven across every directed edge and waited for; and the nodes are compared with ACE — `spock-diff` for configuration and `repset-diff` for data — against a baseline taken before the first node is touched, because the question is whether the upgrade changed anything, not whether the cluster was perfect to begin with.

### DIFF
--- a/.github/workflows/nightly_cluster_upgrade.yml
+++ b/.github/workflows/nightly_cluster_upgrade.yml
@@ -1,0 +1,128 @@
+#
+# Nightly cross-major upgrade of a Spock cluster (no Docker)
+#
+# Builds two PostgreSQL majors plus the Spock extension for each, brings up a
+# three-node Spock mesh on the older one, populates it with the core
+# PostgreSQL regression database, then upgrades the cluster one node at a time
+# and checks that replication survives and the nodes still agree.
+#
+# The whole thing is driven by tests/run-cluster-upgrade.sh, so it runs
+# identically on a developer laptop.  This workflow's job is just: provision
+# deps, invoke the script, keep the evidence.
+#
+# Dependency footprint is dictated by the script's ./configure flags
+# (--enable-debug --enable-cassert --without-icu --without-readline
+# --with-ssl=openssl):
+#
+#   parser/scanner   -> bison, flex
+#   --with-ssl       -> libssl-dev  (also what builds contrib/pgcrypto)
+#   (default zlib)   -> zlib1g-dev
+#   pkg detection    -> pkg-config
+#
+# ICU and readline are switched off on purpose: pg_upgrade compares the locale
+# provider and ICU version between the two clusters, and two independently
+# configured majors are far more likely to agree on plain libc/C.  Everything
+# else the script touches -- gcc, make, perl, git, rsync -- is preinstalled on
+# ubuntu-latest.  If you change the configure flags in the script, sync this
+# list with it.
+#
+# ACE (https://github.com/pgEdge/ace) is checked out and built here because
+# the script uses it to compare the nodes.  Its go.mod asks for a newer Go
+# than the runner image ships, hence the explicit setup-go.
+#
+
+name: Nightly Cluster Upgrade
+run-name: Cross-major cluster upgrade (nightly)
+
+on:
+  workflow_dispatch:          # manual: "Run workflow" button in the UI
+  schedule:
+    # 04:00 UTC, two hours after nightly_tap.yml, so the two do not contend
+    # for runners.  Not hung off pushes: this builds two PostgreSQL majors
+    # from source and takes hours.
+    - cron: '0 4 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  cluster-upgrade:
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+
+    steps:
+      - name: Checkout spock
+        # Full history: the script builds the pre-upgrade nodes from
+        # OLD_SPOCK_REF (v5_STABLE by default) out of this same clone, and a
+        # shallow checkout has no such ref to check out.
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Checkout ACE
+        uses: actions/checkout@v4
+        with:
+          repository: pgEdge/ace
+          path: ace
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update -qq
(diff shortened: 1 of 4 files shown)


samples/Z0DAN/zodan.sql
```diff
@@ -92,6 +92,17 @@ RETURNS int[] LANGUAGE sql IMMUTABLE AS $$
     SELECT (spock.version_to_array(v))[1:2];
 $$;
 
+-- spock.progress's apply-progress column is remote_lsn on Spock 5.x and
+-- remote_commit_lsn on 6.x; resolve it per endpoint.  The IN list keeps %I safe.
+CREATE OR REPLACE FUNCTION spock.progress_commit_lsn_column(p_dsn text)
```
ibrarahmad: LANGUAGE sql means the body is validated at CREATE time, so zodan.sql no longer loads unless dblink is already installed. Checked it against a db with no dblink: main loads clean, this branch stops here with `function dblink(text, unknown) does not exist`. Everything else in the file only touches dblink from plpgsql bodies, which stay lazy, so plpgsql would avoid it.

samples/Z0DAN/zodan.sql
```diff
@@ -2497,10 +2510,12 @@ BEGIN
...
 
-                    SELECT p.remote_commit_lsn INTO target_lsn
-                    FROM spock.progress p
-                    JOIN spock.node n ON n.node_id = p.remote_node_id
-                    WHERE n.node_name = src_node_name;
+                    EXECUTE format(
+                        'SELECT p.%I FROM spock.progress p '
+                        'JOIN spock.node n ON n.node_id = p.remote_node_id '
+                        'WHERE n.node_name = $1',
+                        spock.progress_commit_lsn_column(new_node_dsn))
```
ibrarahmad: This EXECUTE runs in the local session, and add_node already refuses to run anywhere but the new node (1496), so this has the node dblink back to itself just to read its own pg_attribute. If new_node_dsn isn't dialable from the node itself, the WHEN OTHERS at 2540 swallows it and the defensive advance is quietly skipped. A plain local lookup would do here.

samples/Z0DAN/zodan.sql
```diff
@@ -2593,10 +2608,12 @@ BEGIN
...
                 -- that N1 had applied at snapshot time (stored in N3's spock.progress).
-                SELECT p.remote_commit_lsn INTO target_lsn
-                FROM spock.progress p
-                JOIN spock.node n ON n.node_id = p.remote_node_id
-                WHERE n.node_name = rec.node_name;
+                EXECUTE format(
+                    'SELECT p.%I FROM spock.progress p '
+                    'JOIN spock.node n ON n.node_id = p.remote_node_id '
+                    'WHERE n.node_name = $1',
+                    spock.progress_commit_lsn_column(new_node_dsn))
```
ibrarahmad: Same thing, but this one is inside the FOR rec loop, so it's a connection per peer. And the handler at 2656 turns a failure into `Advancing slot ... (error: could not establish connection)`, which kills add_node while pointing at the wrong thing. The src_dsn call at 1903 is fine, that query really is remote.

samples/Z0DAN/zodan.sql
```diff
@@ -1905,12 +1916,14 @@ BEGIN
...
+                        RAISE EXCEPTION 'Cannot read apply progress from source node % (not replication lag): % (SQLSTATE %)',
+                                        src_node_name, SQLERRM, SQLSTATE
+                            USING ERRCODE = SQLSTATE;
                     WHEN OTHERS THEN
-					    -- Let user know if something wrong happens
-						IF verb THEN
-					      RAISE NOTICE 'An error happened: %', SQLERRM;
-						END IF;
                         -- Transient probe failure; restore timeout and retry.
+                        RAISE NOTICE 'An error happened: %', SQLERRM;
```
ibrarahmad: Lost the `IF verb THEN` guard here. The loop retries every 0.5s bounded by spock.sync_timeout (1800s in the upgrade rig), so a peer that's briefly unreachable now floods the client with notices.

ibrarahmad (commented):
Rig itself looks reasonable and the log_min_messages fix in run-single-pg18-installcheck.sh is a good catch on its own. Comments are all on zodan.sql, since that's the part that actually ships.

samples/Z0DAN/zodan.sql
danolivo: @ibrarahmad , the zodan part here has been taken from https://github.com/pgEdge/spock/pull/593 (to avoid waiting until the end of the development cycle). So, I think we can discuss this feature there.

samples/Z0DAN/zodan.sql
danolivo: @ibrarahmad ,
Let's check a fresh version of the branch - I removed any Z0DAN usage from the script. It was my mistake - it can't be used for any major cross-updates, at least for now. Upgrade from PG15 also checked. A draft of the upgrade instruction has been published as a wiki page: https://github.com/pgEdge/spock/wiki/Spock-5-%E2%80%90--6-upgrade-procedure

ibrarahmad (approved):

PR #596, opened 2026-08-27
Resolve unqualified DROP targets before the extension-owned guard.

ibrarahmad:
The guard reads the parse tree, but AutoDDL's hook runs after the statement. For an unqualified DROP the RangeVar carries no schema and the relation is already gone, so `SET search_path TO snowflake; DROP TABLE guard_seed` was replicated while the qualified spelling of it was refused.  The same gap let an unqualified DROP of a node-local pgedge_ace table replicate as well.

spock_ProcessUtility() now resolves the targets while they still exist, and reads no catalog unless it has to.  Covered both ways in 038.

Also label the fenced error block in the release notes, for markdownlint MD040.

### DIFF
--- a/docs/spock_release_notes.md
+++ b/docs/spock_release_notes.md
@@ -277,7 +277,7 @@ AutoDDL has been refactored and hardened:
   statement whose target schema is one of the built-in extension-owned
   schemas now fails outright:
 
-  ```
+  ```text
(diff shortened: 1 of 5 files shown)


mason-sharp:
Reconsidering this (found by Coderabbit), we are making the code more complex when there is a reasonable workaround to simply always namespace/schema-qualify the objects.

PR #597, opened 2026-08-27
Route a reserved extension's tables into a replication set.

ibrarahmad:
DROP EXTENSION cascades away lolor's replication set membership, and the CREATE EXTENSION that follows cannot restore it: autoddl_can_proceed() skips everything nested inside an extension script, so large objects created afterwards were silently not replicated.  Route the member tables of a reserved object with block_in_repset = false, leaving placements be.

#SPOC-669


mason-sharp:
This may be adding some complexity. Couldn't we just document that for lolor if you reinstall it, that you need to add it to a repset again @zaidshabbir25 @ibrarahmad ?

mason-sharp:
Closing this one. It is a similar situation to users wiring up large object handling on their own when they have multiple nodes. Let the user sort through this if they keep creating-dropping-recreating.

PR #602, opened 2026-08-31
Fix apply-group progress entry accounting and initialization

mason-sharp:
Two fixes to `SpockApplyProgress` / apply-group entry handling in shared memory.

### 1. Sync-path allocations wrote uninitialized memory into shmem

Both `SpockApplyProgress` allocations in `spock_sync.c` use plain
`MemoryContextAlloc`, which does not zero, and then set fields by hand.
`adjust_progress_info()` assigns `remote_commit_ts` only when the peer's
column is non-NULL — and the SRF emits NULL whenever that value is 0, which
is the state of every freshly created entry. The garbage is then copied to
`prev_remote_ts` and compared against `last_updated_ts`.

That struct goes to `spock_group_progress_force_set_list()`, which does
`entry->progress = *sap` — the whole struct adopted into shared memory as
authoritative. The invariant fix-ups that follow only ever raise values, so a
garbage LSN that is too high passes straight through. And
`progress_update_struct()` merges monotonically, gated on
`remote_commit_ts`, so one garbage-high value permanently blocks every later
update. Entries are never deleted, so it persists until postmaster restart.

This is not cosmetic: zodan's catchup phase polls `spock.progress` and
proceeds once the LSN passes its target — its own comment warns that going
early risks losing rows in the COPY window — and `spock.lag_tracker` is
computed entirely from these fields.

Both sites now call `spock_init_progress_fields()` right after setting the
key, the helper that is already the single source of truth for "unset".

### 2. `nattached` accounting

Two ways the counter drifted:

- `sub_create()` called `spock_group_attach()` from the client backend just
  to make the entry exist. That backend is not an apply worker and never
  detaches, so every subscription creation permanently incremented the
  count. New `spock_group_ensure_entry()` creates the entry under the master
  lock without claiming attachment; `spock_group_attach()` keeps its one
  legitimate caller.
- The detach ran as `on_proc_exit`, after `spock_worker_on_exit()` releases
  the worker's shmem slot. The manager can hand that slot to a new worker
  within milliseconds, and `spock_worker_attach()` zeroes
  `worker.apply.apply_group` in it, so the exiting process read NULL through
  `MyApplyWorker` and skipped the decrement. Registering it as
  `before_shmem_exit` at attach time fixes the ordering — LIFO runs it while
  the slot is still ours — and pairing it with `spock_apply_worker_attach()`
  also covers a worker that dies before the startup message.

`nattached` is not exposed through any view or function, so neither change
has an observable effect today. The reason to take the second one is the
ordering itself: the old code reached through `MyApplyWorker` into a slot
that may already belong to another worker. The counter is just what happens
to be read there now.

### Testing

None. The initialization bug cannot be covered by a portable regression test
— uninitialized memory that happens to be zero passes any assertion — and
the `nattached` fixes have no SQL-visible effect to assert on. Both were
found and fixed by inspection.

### DIFF
--- a/include/spock_group.h
+++ b/include/spock_group.h
@@ -150,6 +150,8 @@ extern void spock_group_shmem_startup(bool found);
 
 extern SpockGroupEntry *spock_group_attach(Oid dbid, Oid node_id,
 										   Oid remote_node_id);
+extern void spock_group_ensure_entry(Oid dbid, Oid node_id,
+									 Oid remote_node_id);
 extern void spock_group_detach(void);
 extern bool spock_group_progress_update(const SpockApplyProgress *sap);
 extern void spock_group_progress_update_ptr(SpockGroupEntry *entry,
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -2296,7 +2296,6 @@ handle_startup(StringInfo s)
 
 	/* Register callback for cleaning up */
 	before_shmem_exit(spock_apply_worker_shmem_exit, 0);
-	on_proc_exit(spock_apply_worker_on_exit, 0);
 }
 
 /*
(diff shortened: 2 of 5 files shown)


rasifr (approved):

ibrarahmad:
Late note: this also fixed an unlocked HASH_ENTER, since the old sub_create path called spock_group_attach() without holding apply_group_master_lock. attach() still takes no lock of its own and depends on its one caller, which is worth an Assert.

Two nits nearby: spock_sync.c:574 ends in a comma instead of a semicolon, and the comment at spock_functions.c:707 mentions readers of nattached when nothing reads it.

PR #603, opened 2026-09-01
Add test for verifying exception handling- cherry-pick

mason-sharp:
Cherry-picks from PR #585 for v5_STABLE that adds a test for exception handling

### DIFF
--- a/tests/tap/t/039_no_replay_mode_leak.pl
+++ b/tests/tap/t/039_no_replay_mode_leak.pl
@@ -7,6 +7,7 @@ use SpockTest qw(
     get_test_config scalar_query psql_or_bail
     wait_for_sub_status
     log_offset log_since wait_for_log sync_nodes apply_worker_pid
+    poll_query_until
 );
 
 # An ERROR raised in the apply worker BETWEEN two remote transactions sends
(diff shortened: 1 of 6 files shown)


tests/tap/t/039_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,128 @@
...
+    # The transaction that must not be treated as the recorded failure.
+    psql_or_bail(1, "INSERT INTO $t VALUES (2, 'after cancel')");
+    ok(sync_nodes(1, 2),
+        "$mode: worker consumes the stream past the next transaction");
+
+    is(scalar_query(2, "SELECT count(*) FROM $t WHERE id = 2"), '1',
+        "$mode: the next transaction applies");
+
+    unlike(log_since(2, $offset),
+        qr/applied nothing, retrying without exception handling/,
```
danolivo: Does it actually exist in Spock 6?

I see the line:

```
elog(LOG, "SPOCK %s: replay at LSN %X/%X applied nothing ...
```

in v5_STABLE only.

tests/tap/t/039_no_replay_mode_leak.pl
```diff
@@ -0,0 +1,128 @@
...
+    "ARRAY['default', 'default_insert_only', 'ddl_sql'], false, false)");
+ok(wait_for_sub_status(2, 'sub_n1_n2', 'replicating', 30),
+    'subscription starts in replicating state');
+
+for my $phase (@modes) {
+    my $t    = $phase->{table};
+    my $mode = $phase->{mode};
+
+    psql_or_bail(2, "ALTER SYSTEM SET spock.exception_behaviour = $mode");
+    psql_or_bail(2, "SELECT pg_reload_conf()");
```
danolivo, line 84: Being disgustingly neat, I would say this call just sends a signal to the postmaster, not actually changes the value. Hence, races are possible. But I have never seen any actual issues before so that it may be skipped.

danolivo (changes requested):

danolivo (approved):
It looks more or less stable now. Still gaps and races are present, but I think it will happen quite rarely to close our eyes to that now.

PR #606, opened 2026-09-02
Allow zodan add_node to join a Spock 6.0.0 node to a 5.0.9+ cluster

mason-sharp:
Customers on Spock 5.0.9 or later can now add a 6.0.0 node to their cluster with zodan add_node() instead of upgrading every node first. 5.0.9 stays the floor: zodan already refuses anything older, and releases before 5.0.9 have known issues we do not want in production.

Before this change a 6.0.0 node could not join a 5.0.x cluster: zodan required the same major.minor on every node, and even with that gate removed two things broke against a 5.0.x provider.

- The 6.0.0 sync worker calls spock.read_peer_progress() on the provider, which only exists in 6.0.0. Probe the provider's spock_version_num() and, below 6.0.0, read peer progress with an inline query against the 5.x spock.progress table. The resume LSN comes from pg_replication_origin_status in both paths, so its meaning is unchanged.

- zodan polls the source node's spock.progress for remote_commit_lsn, which 5.x still calls remote_lsn. Pick the column name from the source node's Spock version.

Version rule in check_spock_version_compatibility: every node must be 5.0.9 or later; existing nodes must share one major.minor (patch differences allowed); the new node may run the same or a newer major.minor, never an older one, since only the newer sync worker knows how to talk to older providers.

Add tests/tap/t/099_zodan_mixed_version.pl, which adds a current-branch node to a cluster built from an older tag or branch, optionally under pgbench load. Like 014_rolling_upgrade.pl it needs PostgreSQL 18 for extension_control_path.

Tested: 6.0.0 node added to 5.0.9, 5.0.10 and 5.0.11 clusters, idle and under load, plus an all-6.0.0 control run.

### DIFF
--- a/src/spock_sync.c
+++ b/src/spock_sync.c
@@ -622,12 +622,22 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
 	List	   *progress_list = NIL;
 	int			nrows;
 	int			rno;
+	int			provider_version_num = 0;
 
 	/*
-	 * Column indices in the result: lsn(0), snapshot(1) are skipped; GP_*
-	 * start at 2
+	 * Columns (PP_ = peer progress) of the peer progress query below.  Both
+	 * the 6.0.0 form (spock.read_peer_progress) and the pre-6.0.0 form
+	 * (direct read of spock.progress) select exactly these, in this order.
 	 */
-	const int	COL_OFFSET = 2; /* GP_* indices start at COL_OFFSET */
+	enum
+	{
+		PP_REMOTE_NODE_ID = 0,
+		PP_REMOTE_COMMIT_TS,
+		PP_REMOTE_COMMIT_LSN,
+		PP_REMOTE_INSERT_LSN,
+		PP_LAST_UPDATED_TS,
+		PP_UPDATED_BY_DECODE
+	};
 
 	initStringInfo(&query);
 
@@ -721,11 +731,51 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
 	PQclear(res);
 	resetStringInfo(&query);
 
-	/* Read peer progress via the SQL function (slot already exists) */
-	appendStringInfo(&query,
-					 "SELECT * FROM spock.read_peer_progress"
-					 "('%s', %u, %u)",
-					 slot_name, origin_node_id, subscriber_node_id);
+	/*
+	 * Read peer progress: one row per peer subscription on the provider, with
+	 * the PP_ columns declared above.  The resume LSN (remote_commit_lsn) is
+	 * the peer origin's remote_lsn from pg_replication_origin_status, exact
+	 * because apply workers are paused.
+	 *
+	 * A provider older than 6.0.0 has no spock.read_peer_progress() and keeps
+	 * progress in a plain table, so read the same six values from its catalog
+	 * directly.  Detect the version with spock_version_num(), which every
+	 * supported release has.
+	 */
+	res = PQexec(conn, "SELECT spock.spock_version_num()");
+	if (PQresultStatus(res) == PGRES_TUPLES_OK && PQntuples(res) == 1)
+		provider_version_num = atoi(PQgetvalue(res, 0, 0));
+	PQclear(res);
+	elog(LOG, "SPOCK cswp provider spock version_num=%d", provider_version_num);
+
+	if (provider_version_num > 0 && provider_version_num < 60000)
+	{
+		appendStringInfo(&query,
+						 "SELECT p.remote_node_id, p.remote_commit_ts, "
+						 "       COALESCE(ros.remote_lsn, '0/0'::pg_lsn) AS remote_commit_lsn, "
+						 "       p.remote_insert_lsn, p.last_updated_ts, p.updated_by_decode "
+						 "FROM spock.subscription sub "
+						 "JOIN spock.progress p ON p.remote_node_id = sub.sub_origin "
+						 "                     AND p.node_id = sub.sub_target "
+						 "JOIN pg_replication_origin o ON o.roname = sub.sub_slot_name "
+						 "LEFT JOIN pg_replication_origin_status ros ON ros.local_id = o.roident "
+						 "WHERE sub.sub_target = %u AND sub.sub_origin <> %u",
+						 origin_node_id, subscriber_node_id);
+	}
+	else
+	{
+		/*
+		 * read_peer_progress() also returns a header row (remote_node_id
+		 * NULL) and lsn/snapshot columns; neither is needed here because the
+		 * slot's LSN and snapshot were taken from CREATE_REPLICATION_SLOT.
+		 */
+		appendStringInfo(&query,
+						 "SELECT remote_node_id, remote_commit_ts, remote_commit_lsn, "
+						 "       remote_insert_lsn, last_updated_ts, updated_by_decode "
+						 "FROM spock.read_peer_progress('%s', %u, %u) "
+						 "WHERE remote_node_id IS NOT NULL",
+						 slot_name, origin_node_id, subscriber_node_id);
+	}
 	res = PQexec(conn, query.data);
 	resetStringInfo(&query);
 
@@ -733,20 +783,12 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
 		elog(ERROR, "could not create replication slot and get progress on "
 			 "origin: %s", PQerrorMessage(conn));
 
+	/* Zero rows is legitimate: the provider may have no other peers. */
 	nrows = PQntuples(res);
-	if (nrows < 1)
-		elog(ERROR, "spock.read_peer_progress returned no rows");
-
-	/*
-	 * Row 0 is the header row: lsn + snapshot, progress fields all NULL.
-	 * lsn_out and snapshot are already set from the replication protocol;
-	 * just log for debugging.  Skip COL_LSN/COL_SNAP from SQL result.
-	 */
 	elog(LOG, "SPOCK cswp slot=%s lsn=%X/%X snapshot=%s peers=%d",
-		 slot_name, LSN_FORMAT_ARGS(*lsn_out), snapshot, nrows - 1);
+		 slot_name, LSN_FORMAT_ARGS(*lsn_out), snapshot, nrows);
 
-	/* Rows 1+ are progress entries (remote_node_id NOT NULL) */
-	for (rno = 1; rno < nrows; rno++)
+	for (rno = 0; rno < nrows; rno++)
 	{
 		SpockApplyProgress *sap;
 		MemoryContext oldctx;
@@ -756,7 +798,7 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
 		char	   *remote_insert_lsn_str;
 		char	   *last_updated_ts_str;
 
-		if (PQgetisnull(res, rno, COL_OFFSET + GP_REMOTE_NODE_ID))
+		if (PQgetisnull(res, rno, PP_REMOTE_NODE_ID))
 			continue;			/* shouldn't happen but be safe */
 
 		sap = (SpockApplyProgress *) MemoryContextAlloc(CacheMemoryContext,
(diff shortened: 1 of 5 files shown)


ibrarahmad:
`adjust_progress_info()` (src/spock_sync.c:471) still reads the provider's `spock.progress` using the 6.0 `GP_*` ordinals, but on a 5.x provider that query returns 7 columns starting at `node_id`, not 10 starting at `dboid`, so the fields are read from the wrong columns. `copy_tables_data()` calls it unconditionally, so a single table resync from a 5.x provider breaks.

The 6.0 detection uses `spock.spock_version_num()`, which is the library version, but what it needs to know is whether `spock.read_peer_progress()` exists. A provider on the 6.0 binary that has not had `ALTER EXTENSION spock UPDATE` run reports 60000 and then fails on the missing function. The probe also runs after `SET TRANSACTION SNAPSHOT`, so if it ever does fail the transaction is aborted and `resume_apply_workers()` never runs.

`099_zodan_mixed_version` is not listed in tests/tap/schedule or schedule-nightly.

src/spock_sync.c
```diff
@@ -721,32 +731,64 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
...
+	 * the PP_ columns declared above.  The resume LSN (remote_commit_lsn) is
+	 * the peer origin's remote_lsn from pg_replication_origin_status, exact
+	 * because apply workers are paused.
+	 *
+	 * A provider older than 6.0.0 has no spock.read_peer_progress() and keeps
+	 * progress in a plain table, so read the same six values from its catalog
+	 * directly.  Detect the version with spock_version_num(), which every
+	 * supported release has.
+	 */
+	res = PQexec(conn, "SELECT spock.spock_version_num()");
```
ibrarahmad: This is the library version, but what the branch below needs to know is whether `spock.read_peer_progress()` exists. A provider running the 6.0 binary without `ALTER EXTENSION spock UPDATE` reports 60000, takes the 6.0 path, and then fails on the missing function, which is exactly the state a rolling upgrade passes through. `to_regprocedure('spock.read_peer_progress(text,oid,oid)') IS NOT NULL`, or `extversion` from `pg_extension` as `check_spock_version_compatibility` already does, would answer the actual question.

Separately, the probe runs after the `SET TRANSACTION SNAPSHOT` above, so the tolerated failure is not really tolerable: a failed statement aborts the transaction, the next `PQexec` reports "current transaction is aborted" under the "could not create replication slot and get progress" message, and `resume_apply_workers()` is never reached. Moving it above `pause_apply_workers()`, next to the existing `PQserverVersion(conn)` check, fixes both.

src/spock_sync.c
```diff
@@ -721,32 +731,64 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
...
+	PQclear(res);
+	elog(LOG, "SPOCK cswp provider spock version_num=%d", provider_version_num);
+
+	if (provider_version_num > 0 && provider_version_num < 60000)
+	{
+		appendStringInfo(&query,
+						 "SELECT p.remote_node_id, p.remote_commit_ts, "
+						 "       COALESCE(ros.remote_lsn, '0/0'::pg_lsn) AS remote_commit_lsn, "
+						 "       p.remote_insert_lsn, p.last_updated_ts, p.updated_by_decode "
+						 "FROM spock.subscription sub "
```
ibrarahmad: `adjust_progress_info()` at line 471 is the other place that reads a provider's `spock.progress`, and it is still version blind: it does `SELECT * FROM spock.progress` on `origin_conn` and indexes the result with the `GP_*` ordinals from spock_group.h, which describe 6.0's 10 column `apply_group_progress()` view. A 5.x provider returns 7 columns starting at `node_id`, so `GP_REMOTE_COMMIT_LSN` lands on `last_updated_ts` and `GP_UPDATED_BY_DECODE` is past the end of the result. `copy_tables_data()` calls it unconditionally, so a single table resync from a 5.x provider breaks, and that is reachable in normal operation once mixed clusters are allowed.

Related to the query here: on 5.x `spock.progress` is a table flushed at most once per second by `flush_progress_if_needed()`, not a view over shared memory, so a peer that has not flushed yet has no row and the inner join drops it silently. The `nrows < 1` guard that used to catch an empty result is gone too. A `LEFT JOIN` from `spock.subscription` with coalesced defaults would keep the peer list complete.

tests/tap/t/099_zodan_mixed_version.pl
```diff
@@ -0,0 +1,315 @@
+use strict;
```
ibrarahmad: This is not listed in tests/tap/schedule or tests/tap/schedule-nightly, so it never runs. Worth adding to nightly, or a comment saying it is manual only the way 018_upgrade_schema_match has one.

Two other things: the PR description says PostgreSQL 18 is needed for `extension_control_path`, but this test copies the whole install and relies on paths being resolved relative to `bin/`, which is what the header comment says as well, so the description looks stale. And the version rule itself has no coverage here: an older new node being rejected, and existing nodes with differing major.minor being rejected, are the actual behaviour changes in `check_spock_version_compatibility`.

src/spock_sync.c
ibrarahmad: This is the library version, but the branch below needs to know whether `spock.read_peer_progress()` exists. A provider on the 6.0 binary that has not had `ALTER EXTENSION spock UPDATE` run reports 60000 and then fails on the missing function. `to_regprocedure('spock.read_peer_progress(text,oid,oid)') IS NOT NULL` answers the real question.

It also runs after `SET TRANSACTION SNAPSHOT`, so a failure here aborts the transaction and `resume_apply_workers()` is never reached. Probing above `pause_apply_workers()` avoids that.

src/spock_sync.c
```diff
@@ -721,32 +731,64 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
...
+	res = PQexec(conn, "SELECT spock.spock_version_num()");
+	if (PQresultStatus(res) == PGRES_TUPLES_OK && PQntuples(res) == 1)
+		provider_version_num = atoi(PQgetvalue(res, 0, 0));
+	PQclear(res);
+	elog(LOG, "SPOCK cswp provider spock version_num=%d", provider_version_num);
+
+	if (provider_version_num > 0 && provider_version_num < 60000)
+	{
+		appendStringInfo(&query,
+						 "SELECT p.remote_node_id, p.remote_commit_ts, "
```
ibrarahmad, line 792: `adjust_progress_info()` at line 471 is the other reader of a provider's `spock.progress` and is still version blind. It indexes `SELECT *` with the `GP_*` ordinals for 6.0's 10 column view, while a 5.x provider returns 7 columns starting at `node_id`. `copy_tables_data()` calls it unconditionally, so a single table resync from a 5.x provider reads the wrong fields.

src/spock_sync.c
```diff
@@ -721,32 +731,64 @@ spock_create_slot_and_read_progress(PGconn *conn, PGconn *repl_conn,
...
+						 "       COALESCE(ros.remote_lsn, '0/0'::pg_lsn) AS remote_commit_lsn, "
+						 "       p.remote_insert_lsn, p.last_updated_ts, p.updated_by_decode "
+						 "FROM spock.subscription sub "
+						 "JOIN spock.progress p ON p.remote_node_id = sub.sub_origin "
```
ibrarahmad, line 797: On a 5.x provider `spock.progress` is a table that `flush_progress_if_needed()` writes at most once per second, not a view over shared memory, so `p.remote_commit_ts`, `p.remote_insert_lsn` and `p.last_updated_ts` can lag the paused state here. The resume LSN comes from `ros` and stays exact, but the comment above is only true of that one column on this path.

tests/tap/t/099_zodan_mixed_version.pl
ibrarahmad, line 1: Not listed in tests/tap/schedule or schedule-nightly, so it never runs. The version rule itself is also uncovered: rejecting an older new node, and rejecting existing nodes with differing major.minor, are the actual behaviour changes in `check_spock_version_compatibility`.

src/spock_sync.c
mason-sharp: Good catch, on both counts. Fixed in d98031cd.

The probe is now to_regprocedure('spock.read_peer_progress(text,oid,oid)') IS NOT NULL, so it asks the catalog what actually matters. A provider on the 6.0 binary with ALTER EXTENSION spock UPDATE still pending takes the legacy path, which is correct: in that state spock.progress is still the 5.x table.

The probe also moved above pause_apply_workers(). It does not depend on the snapshot, and any failure there now happens before the workers are paused. For what it's worth,
both counts. Fixed in d98031cd.

The probe is now to_regprocedure('spock.read_peer_progress(text,oid,oid)') IS NOT NULL, so it asks the catalog what actually matters. A provider on the 6.0 binary with ALTER EXTENSION spock UPDATE still pending takes the legacy path, which is correct: in that state spock.progress is still the 5.x table.

The probe also moved above pause_apply_workers(). It does not depend on the snapshot, and any failure there now happens before the workers are paused. For what it's worth, the exposure before this fix was bounded: the apply workers' pause loop gives up after spock_pause_timeout and clears the flag itself. But there is no reason to spend that window on a probe.

tests/tap/t/099_zodan_mixed_version.pl
mason-sharp: Correct, and the comment over-claimed. Fixed in d98031cd.

The comment now says that on a pre-6.0.0 provider only the resume LSN is exact, and that remote_commit_ts, remote_insert_lsn, last_updated_ts and updated_by_decode can trail the paused state by up to the one-second flush interval. That is acceptable because those values only seed the new node's lag tracking and are replaced by the first replicated message.

I also clamped remote_insert_lsn to GREATEST(p.remote_insert_lsn, ros.remote_lsn) on that path. Without it, the new node's lag_tracker can show a negative byte lag for up to a second after the add, and the earlier test logs showed exactly that (commit_lsn=0/2FE0848 insert_lsn=0/2FE07C8).

src/spock_sync.c
mason-sharp: You're right, and it was worse than blind. On a 5.x provider SELECT * returns seven columns starting at node_id, so GP_REMOTE_NODE_ID (ordinal 2) would have read remote_commit_ts as a node id. The test never caught it because add_node goes through the slot path and never reaches copy_tables_data().

Fixed in d98031cd. adjust_progress_info() now checks whether the provider's spock.progress has a remote_commit_lsn column and selects the six columns by name, remote_lsn or remote_commit_lsn as appropriate. Both readers now share one column list, one enum and one row parser, so there are no GP_* ordinals left in spock_sync.c and no second place for this to drift.

The test now also runs a single-table resync from the 5.0.11 provider after the add, checks that adjust_progress_info() ran, and compares the table on both nodes.

tests/tap/t/099_zodan_mixed_version.pl
```diff
@@ -0,0 +1,360 @@
...
+# so it covers the other place the sync worker must understand an older
+# provider's catalog.
+psql_or_bail(1, "UPDATE pgbench_branches SET bbalance = bbalance + 1");
+psql_or_bail(1, 'SELECT spock.wait_slot_confirm_lsn(NULL, NULL)');
+psql_or_bail(3, "SELECT spock.sub_resync_table('sub_n1_n3', 'pgbench_branches')");
+wait_until(3, "SELECT status FROM spock.sub_show_table('sub_n1_n3', 'pgbench_branches')", 'replicating', 120,
+    "single-table resync of pgbench_branches on n3 from n1 ($v[0]) completed");
+my $n3_log = "$ENV{TESTLOGDIR}/00$ports->[2].log";
+my $adjusted = `grep -c 'SPOCK: adjust spock.progress' '$n3_log' 2>/dev/null`; chomp $adjusted;
+cmp_ok($adjusted, '>=', 1, "resync read the provider's progress entries (adjust_progress_info ran)");
```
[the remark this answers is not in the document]
mason-sharp: Taken, in the form proposed: the check now records the count before the resync and asserts it increased.

One correction to the premise. adjust_progress_info() does not run during add_node. The initial subscription sync reads peer progress in spock_create_slot_and_read_progress() and then calls copy_replication_sets_data() with progress_out = NULL, which skips adjust_progress_info(). Only copy_tables_data(), the single-table resync path, calls it unconditionally. The n3 log has zero adjust spock.progress lines before the resync, so the original >= 1 assertion was not vacuous.

The baseline form is still the right one: the same log prefix is also written by spock_group_progress_update_list(), and a future change to the add path should not be able to mask a regression here.

ibrarahmad (approved):

PR #612, opened 2026-09-10
apply: handle clock skew in progress tracking

rasifr:
Origin commit timestamps may be ahead of the subscriber clock and, for forwarded transactions, need not increase with the provider WAL position.

Remove the assertion that local apply time follows origin commit time. Advance remote_commit_lsn independently of the timestamp fields, while keeping remote_commit_ts and its local observation fields paired for lag reporting.

Treat prev_remote_ts as a commit-order synchronization token rather than a timestamp high-water mark. Update it for commit records even when their timestamp is below the recorded maximum, while leaving it unchanged for statistics-only updates.

### DIFF
--- a/src/spock_apply.c
+++ b/src/spock_apply.c
@@ -1170,8 +1170,10 @@ handle_commit(StringInfo s)
 			.updated_by_decode = true,
 		};
 
-		/* XXX: Don't care in production yet */
-		Assert(sap.last_updated_ts >= sap.remote_commit_ts);
+		/*
+		 * The origin commit time and local apply time come from independent
+		 * clocks, so no ordering relationship is required between them.
+		 */
 
 		Assert(MyApplyWorker && MyApplyWorker->apply_group);
 
--- a/src/spock_group.c
+++ b/src/spock_group.c
@@ -316,22 +316,32 @@ progress_update_struct(SpockApplyProgress *dest, const SpockApplyProgress *src)
 	Assert(dest->key.node_id == src->key.node_id);
 	Assert(dest->key.remote_node_id == src->key.remote_node_id);
 
+	/*
+	 * Track the commit position independently of commit timestamps. Forwarded
+	 * transactions retain the original node's timestamp, so timestamp order
+	 * need not match the provider's WAL order.
+	 *
+	 * Keep remote_commit_ts and its local observation fields together so lag
+	 * reporting uses values from the same commit. The resulting timestamp may
+	 * describe an earlier commit than remote_commit_lsn.
+	 *
+	 * prev_remote_ts is the synchronization token consumed by parallel apply,
+	 * not a high-water mark. Update it for commit records, including commits
+	 * whose timestamps are below the recorded maximum. Statistics-only
+	 * updates have remote_commit_ts == 0 and must leave it unchanged.
+	 */
+	if (dest->remote_commit_lsn < src->remote_commit_lsn)
+		dest->remote_commit_lsn = src->remote_commit_lsn;
+
 	if (dest->remote_commit_ts < src->remote_commit_ts)
 	{
-		/*
-		 * This is the most advanced commit. Save its progress.
-		 *
-		 * NOTE: According to apply group machinery their commit order should
-		 * follow the timestamp order. That means there are no way for a
-		 * commit to come with an oldest commit timestamp except we don't
-		 * update this commit's part of the data at all.
-		 */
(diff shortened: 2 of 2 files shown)


src/spock_group.c
```diff
@@ -316,23 +316,33 @@ progress_update_struct(SpockApplyProgress *dest, const SpockApplyProgress *src)
...
+	 * Keep remote_commit_ts and its local observation fields together so lag
+	 * reporting uses values from the same commit. The resulting timestamp may
+	 * describe an earlier commit than remote_commit_lsn.
+	 *
+	 * prev_remote_ts is the synchronization token consumed by parallel apply,
+	 * not a high-water mark. Update it for commit records, including commits
+	 * whose timestamps are below the recorded maximum. Statistics-only
+	 * updates have remote_commit_ts == 0 and must leave it unchanged.
+	 */
+	if (dest->remote_commit_lsn < src->remote_commit_lsn)
```
mason-sharp, line 333: With the LSN now advancing on its own, remote_commit_ts and remote_commit_lsn in spock.progress can describe two different commits, as the comment above says. The struct doc still promises the opposite: include/spock_group.h:102 reads "remote_commit_lsn - LSN of the COMMIT corresponding to the remote_commit_ts." Could you update that line in this PR (something like "the most advanced COMMIT LSN seen on this stream; may belong to a later commit than remote_commit_ts when transactions are forwarded")?

src/spock_group.c
```diff
@@ -316,23 +316,33 @@ progress_update_struct(SpockApplyProgress *dest, const SpockApplyProgress *src)
...
-		 * update this commit's part of the data at all.
-		 */
 		dest->remote_commit_ts = src->remote_commit_ts;
-		dest->prev_remote_ts = src->prev_remote_ts;
-		dest->remote_commit_lsn = src->remote_commit_lsn;
 		dest->last_updated_ts = src->last_updated_ts;
 		dest->updated_by_decode = src->updated_by_decode;
 	}
 
+	if (src->remote_commit_ts != 0)
```
mason-sharp, line 343: This makes prev_remote_ts follow the last commit record, which is the right thing for wait_for_previous_transaction() (equality test, spock_apply.c:311). One consumer still treats it as a high-water mark, though: spock_progress_recovery.c:359 rebuilds it as running_max_ts after a crash. With forwarding, a follower whose predecessor was a low-ts forwarded commit could then wait on a value the recovered entry never takes. Possibly out of scope here, but worth a comment at the recovery site, or a follow-up issue, so the two readings of the field don't drift apart again.

src/spock_apply.c
```diff
@@ -1170,8 +1170,10 @@ handle_commit(StringInfo s)
 			.updated_by_decode = true,
 		};
 
-		/* XXX: Don't care in production yet */
-		Assert(sap.last_updated_ts >= sap.remote_commit_ts);
+		/*
+		 * The origin commit time and local apply time come from independent
```
mason-sharp: Agree with dropping the Assert. 

Say which clocks these are, so nobody reinstates it: remote_commit_ts is the origin's commit timestamp (logical-clock bumped on the patched server), last_updated_ts is this node's system clock at apply time; the only comparisons that mean anything are commit-ts against commit-ts, or a local reading against the same local clock.

FYI, I have a separate branch I am working on that adds more comments about *_commit_ts being from HLC on the origins, and the others (usually) being wall clock timestamps, but still may be worth a brief mention here while changing.

mason-sharp (changes requested):
Some feedback. Also try to include tests if possible.

mason-sharp:
Can we add a test here? One TAP test, say 047_progress_clock_skew.pl, reusing the 3-node topology from 015_forward_origin_advance.pl (n2 → n1 → n3, forward_origins='all' on n3's sub to n1). Forwarding gives you ts/LSN divergence without needing to touch any clock. Four assertions:

- remote_commit_lsn is non-decreasing across samples from spock.progress on n3 even across samples where remote_commit_ts moves backward. That's change #1, observable purely in SQL.
- prev_remote_ts can sit below remote_commit_ts — previously impossible, so it's a positive check that change #2 took effect.
- n3's row count converges within a bounded timeout. This is the real assertion: if prev_remote_ts is wrong, wait_for_previous_transaction() spins forever and you get a timeout rather than a wrong value.
- SIGKILL n3, restart, assert prev_remote_ts IS NULL (the new documented behavior) and that replication resumes and converges.

mason-sharp (approved):

